PR #408, opened 2023-01-05
vuu-filters types

harryhartley:


vuu-ui/packages/vuu-filters/src/filter-toolbar/useFilterToolbar.tsx
```diff
@@ -35,15 +31,16 @@ const filterToControl = (
       >
         <FilterDropdown
           column={column}
-          selected={value}
+          selected={value.toString()}
```
heswell, line 34: This is ok . I think to fully capture the data type FilterDropdown should maybe be generic, just as the nested Dropdown is (defaults to Dropdown<string>. Gets complicated though given that this is highly dynamic and may be a number or string. toString on everything is the most straightforward option

vuu-ui/packages/vuu-filters/src/filter-toolbar/useFilterToolbar.tsx
```diff
@@ -52,21 +49,20 @@ const filterToControl = (
         key={column}
         labelPlacement="top"
       >
-        <FilterDropdown
+        { values.map((value, idx) => <FilterDropdown
```
heswell, line 52: This looks a bit odd. Instead os a single multi select dropdown that will show the filter values, we have multiple dropdowns each showing one value, was this intended ?

heswell (commented):
Great work. One issue with the FilterDropdown piece. 

This stuff around the FilterToolbar was highly experimental. Do with it whatever you like - I imagine it might be discarded entirely once you start work on newer filtering ui.

vuu-ui/packages/vuu-filters/src/filter-toolbar/useFilterToolbar.tsx
harryhartley: Agree that this is incorrect, I think I misunderstood what was happening with this component - rolling back the change

harryhartley (commented):
Addressed comments, rolled back filter dropdown change

PR #424, opened 2023-01-12
Typescript uplift layout package

harryhartley:
Bit of (opinionated) tidying of the layout package - reviewed by Luke internally, happy to rollback any changes you're not happy with (notably the tidying of most comments, especially those over a year old)


vuu-ui/packages/vuu-layout/src/drag-drop/Draggable.ts
```diff
@@ -216,24 +216,24 @@ function dragMousemoveHandler(evt: MouseEvent) {
     _dragMoveCallback?.(newX, newY);
   }
 
-  if (_simpleDrag) {
+  if (_simpleDrag || !_dragContainer || !_validDropTargetPaths) {
```
heswell, line 219: early exit in the case of !_validDropTargetPaths is a breaking change. It's normal for 
_validDropTargetPaths to not be defined. The use of a non-null assertion on the original code was wrong.

vuu-ui/packages/vuu-layout/src/flexbox/Flexbox.tsx
```diff
@@ -10,12 +9,10 @@ const classBase = "hwFlexbox";
 
 const Flexbox = forwardRef(function Flexbox(
   props: FlexboxProps,
-  ref: ForwardedRef<HTMLDivElement>
```
heswell, line 13: the forwardedRef here isn't strictly needed - it's a courtesy for users who might want to capture a ref to the root element. If we DID want to remove the forwardedRef, though, the forwardRef call against the component should be removed as well.

vuu-ui/packages/vuu-layout/src/flexbox/Flexbox.tsx
```diff
@@ -49,10 +45,8 @@ const Flexbox = forwardRef(function Flexbox(
     <div
       {...rest}
       className={className}
-      // data-cols={cols}
       data-resizeable={resizeable || undefined}
       id={id}
-      ref={useForkRef(rootRef, ref)}
```
heswell, line 55: this one is needed, removing it breaks drag drop functionality. I've reverted this change and fixed the ref typing in the hook return type.

harryhartley:
Thanks for fixing these @heswell, appreciate the additional context!

PR #425, opened 2023-01-13
Typescript uplift layout package

heswell:
This is 98% the changes from  PR #424 with a couple of amendments. I also deleted some unused files from layout and added layout to the list of packages we generate type-defs for when publishing.


vuu-ui/packages/vuu-layout/src/layout-reducer/layoutUtils.ts
```diff
@@ -243,9 +225,7 @@ export function layoutFromJson(
     componentType,
     {
       ...props,
-      id,
```
heswell, line 246: id is important. Fix is to make sure componentType rexognizes id.

vuu-ui/packages/vuu-layout/src/layout-reducer/layoutUtils.ts
```diff
@@ -243,9 +225,7 @@ export function layoutFromJson(
...
       key: id,
-      path,
```
heswell, line 248: likewise, path is critical for functioning of layout drag drop

PR #451, opened 2023-01-23
Update README to include IntelliJ setup

harryhartley:


harryhartley:
Hey Steve, just a little change documenting Luke's experiments with IntelliJ, thought it might be more appropriate here - happy to rewrite anything that's unclear

README.md
```diff
@@ -35,6 +35,26 @@ Windows:
...
+3. Set project SDK version to 11: file -> project structure -> select an SDK -> require version 11
+4. Enable 'Use plugin registry': file -> settings -> build, execution, deployment -> Maven
+5. Open Maven tab on the right and click install on vuu-parent -> lifecycle -> install
+6. In the terminal, navigate to 
+```sh
+vuu-ui/sample-apps/app-vuu-example
+```
+7. Run 
+```sh
+yarn build
```
heswell, line 53: this is the same as the
```
yarn:build app
```
line in the client instructions below. This isn't wrong, of course, but won't work unless the user has done an initial 
```
yarn
```
to install everything. Probably best to just instruct user to follow the client installation below before running the vuu server for first time,.

harryhartley: I've added the instruction and moved it below client installation so it makes chronological sense

PR #506, opened 2023-02-10
Issue 502 menu filters

heswell:
add support for filter evaluation when building context menus

add support for nested filters on cliuent

add field to CellMenu so cell-scoped MenuItem can be rendered on targeted column

### DIFF
--- a/vuu/src/main/scala/org/finos/vuu/core/module/simul/SimulationModule.scala
+++ b/vuu/src/main/scala/org/finos/vuu/core/module/simul/SimulationModule.scala
@@ -77,15 +77,28 @@ class InstrumentsService(val table: DataTable, val providerContainer: ProviderCo
     NoAction()
   }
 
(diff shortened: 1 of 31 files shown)


vuu/src/main/scala/org/finos/vuu/core/module/simul/SimulationModule.scala
```diff
@@ -77,13 +77,28 @@ class InstrumentsService(val table: DataTable, val providerContainer: ProviderCo
     NoAction()
   }
 
-  override def menuItems(): ViewPortMenu = ViewPortMenu("Test Menu",
+  override def menuItems(): ViewPortMenu = ViewPortMenu(
+    ViewPortMenu("Block Actions (selected rows)",
```
chrisjstevo, line 81: Why so many quotes on the filter definition..?

heswell: fixed

Issue #511, opened 2023-02-13
Latest UI code expects server to support quotes around all string values in filters 


chrisjstevo:
## Bug Report

The UI code has recently changed from defining strings as literals without quotes to literals with quotes. 

This has meant that previous search strings like: 
```ric starts A ```

has become:
```ric starts "A"```

These quote marks are already stripped out for = case ( ric = "FOO") but they now need to be stripped out for any string operation. 

[relevant test](https://github.com/finos/vuu/blob/main/vuu/src/test/scala/org/finos/vuu/core/filter/FilterGrammerTest.scala)
[EqualsClause](https://github.com/finos/vuu/blob/main/vuu/src/main/scala/org/finos/vuu/core/filter/FilterClause.scala) which stripQuotes() when is a string

In the same file with have "starts, ends, in [ ]" clauses which do not stripQuotes. Should be trivial to add. 

Tests cases we need to add are: 

```
field starts "X"
field ends "X"
field in ["X", "Y"]
field != "X"
```

chrisjstevo:
Hi. This is stripping quotes when we're doing the comparison. When the token is handed to the scala code its handed with quotes: so we get a string value that = "\"FOO\"" we want the string to be "FOO" so that we can do a comparison

heswell:
I see - antlr correctly parses it as a string, but gives it as a token , quotes included. Makes sense.

PR #519, opened 2023-02-15
Latest UI code expects server to support quotes around all string val…

keikeicheung:
…ues in filters #511

- rearranged the tests a bit for test run window able to show all tests
- remove one dup test - ric in [AAPL.L,BT.L]


chrisjstevo:
Looks good. :-)

heswell:
Nice work

PR #550, opened 2023-03-01
Filters improvements

ddimtirov:
Reviewed internally in UBS (MR !19)

### DIFF
--- a/vuu/src/main/scala/org/finos/vuu/core/filter/FilterTreeVisitor.scala
+++ b/vuu/src/main/scala/org/finos/vuu/core/filter/FilterTreeVisitor.scala
@@ -1,7 +1,7 @@
 package org.finos.vuu.core.filter
 
 import com.typesafe.scalalogging.StrictLogging
-import org.finos.vuu.grammar.FilterBaseVisitor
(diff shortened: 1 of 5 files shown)


vuu/src/main/scala/org/finos/vuu/core/filter/FilterTreeVisitor.scala
```diff
@@ -70,12 +70,26 @@ class FilterTreeVisitor extends FilterBaseVisitor[FilterClause] with StrictLoggi
...
     val atomType = ctx.head.getChild(0).asInstanceOf[TerminalNode].getSymbol.getType
-    (atomType, ctx.map(ctx => ctx.getText))
+    val textValues = ctx.map(ctx => parseValue(atomType, ctx))
+
+    (atomType, textValues)
+  }
+
+  private def parseValue(atomType: Int, ctx: AtomContext) = {
+    atomType match {
+      case FilterLexer.STRING =>
```
heswell, line 87: Would be nice if we could leave this to the parser - define a token for the string content (without the quotes) and define the grammar for a String as something like
```
'"' StringContent '"'
```
. Not sure if we can do it in such a way that it would not clash with the token for identifier, though.

PR #552, opened 2023-03-01
logging in dev mode, no logging in prod

theiceoan:
Logging works when in dev mode (build is run in dev mode with --dev flag)
Logger class created that supports all the functions of console
All console.log statements have been replaced and a bit of additional logging has been added to the vuu-data files

### DIFF
--- a/vuu-ui/packages/vuu-data/src/server-proxy/server-proxy.ts
+++ b/vuu-ui/packages/vuu-data/src/server-proxy/server-proxy.ts
@@ -452,51 +452,37 @@ export class ServerProxy {
         }
       } else {
         const viewport = this.getViewportForClient(message.viewport);
+        logger.log(`${message.type} Message From Client: `, message);
         switch (message.type) {
           case "setViewRange":
-            logger.log('Set View-Range Message From Client: ', message);
             return this.setViewRange(viewport, message);
           case "aggregate":
-            logger.log('Aggregate Message From Client: ', message);
             return this.aggregate(viewport, message);
           case "sort":
-            logger.log('Sort Message From Client: ', message);
             return this.sort(viewport, message);
           case "groupBy":
-            logger.log('GroupBy Message From Client: ', message);
             return this.groupBy(viewport, message);
           case "filter":
-            logger.log('Filter Message From Client: ', message);
             return this.filter(viewport, message);
           case "select":
-            logger.log('Select Message From Client: ', message);
             return this.select(viewport, message);
           case "suspend":
-            logger.log('Suspend Message From Client: ', message);
             return viewport.suspend();
           case "resume":
-            logger.log('Resume Message From Client: ', message);
             return this.resumeViewport(viewport);
           case "enable":
-            logger.log('Enable Message From Client: ', message);
             return this.enableViewport(viewport);
           case "openTreeNode":
-            logger.log('Open Tree Node Message From Client: ', message);
             return this.openTreeNode(viewport, message);
           case "closeTreeNode":
-            logger.log('Close Tree Node Message From Client: ', message);
             return this.closeTreeNode(viewport, message);
           case "createLink":
-            logger.log('Create Link Message From Client: ', message);
             return this.createLink(viewport, message);
           case "removeLink":
-            logger.log('Remove Link Message From Client: ', message);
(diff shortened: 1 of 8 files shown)


vuu-ui/packages/vuu-utils/src/logging-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+	assert: AssertLogFn;
+	trace: LogFn;
+	debug: LogFn;
+	info: LogFn;
+	table: TableLogFn
+}
+
+export type BuildEnv = 'production' | 'development';
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const NO_OP: LogFn = () => {};
```
heswell, line 28: if you want to avoid the eslint dodging, just define the function as
```
() => undefined
```

vuu-ui/packages/vuu-utils/src/logging-level.ts
```diff
@@ -0,0 +1,5 @@
+// initial code for the use of cookies for logging level
```
heswell, line 1: any particular reason this lives in a file of its own ? Feels like it might as well go in  logging-utils

vuu-ui/packages/vuu-utils/src/logging-level.ts
```diff
@@ -0,0 +1,5 @@
...
+export const loggingLevel = (level:string) => {
+	const date = new Date();
```
heswell, line 3: where would this be called from ?

vuu-ui/packages/vuu-utils/src/logging-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+	readonly assert: AssertLogFn;
+	readonly trace: LogFn;
+	readonly debug: LogFn;
+	readonly info: LogFn;
+	readonly table: TableLogFn;
+
+	constructor(options?: { buildEnv?: string }, level?: string) {
+		const { buildEnv } = options || {};
+
+		if (buildEnv === 'production') {
```
heswell, line 46: I'm not sure about this approach tbh.  Firstly, even having NO_OP method calls in performance sensitive sections a code worries me a bit. It may be that the javascript runtime is smart enough to optimise them away entirely, maybe not. But then any runtime arguments are definitiely going to be evaluated, whether the function does anything or not. 
There may be cases where we want to build a complex structured message, maybe just a JSON .stringify - in these cases we will still be occurring that cost even when the log method is a noop.

What I was thinking was that we would wrap the actual logging statements with environment checks, using substituted environment variables, so that in a prod build, there is no logging code at all in the source (the minifier would eliminate it)

vuu-ui/packages/vuu-data/src/server-proxy/server-proxy.ts
```diff
@@ -451,34 +454,49 @@ export class ServerProxy {
         const viewport = this.getViewportForClient(message.viewport);
         switch (message.type) {
           case "setViewRange":
+            logger.log('Set View-Range Message From Client: ', message);
```
heswell, line 457: this feels like overkill. Why wouldn't we just log one message,
```
logger.log(`${message.type} message from client`)
```

vuu-ui/packages/vuu-utils/src/logging-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+		this.groupCollapsed = console.groupCollapsed.bind(console);
+		this.assert = console.assert.bind(console);
+		this.trace = console.trace.bind(console);
+		this.debug = console.debug.bind(console);
+		this.info = console.info.bind(console);
+		this.table = console.table.bind(console);
+	}
+}
+
+export const logger = new ConsoleLogger({ buildEnv: process.env.NODE_ENV });
```
heswell, line 76: I think there are advantages to letting each call site define a logger, that way they can add context. We're not really adding a lot of value to console.log lijke this

so for example, in ServerProxy,.ts, I might have this
```
import {Logger} from "@finos/utils"

const logger = new Logger('ServerProxy')
```
then within the code, we use it exactly the way you already have

we can then format the log messages, so they appear something like

```bash
>> [ServerProxy] : 10:11:25.000000 message provided by user
```

PR #569, opened 2023-03-08
Issue 542 logging in vuu-data package

theiceoan:
Implementation of logging levels.
There are 3 logging levels: high, medium, low 
These can be manipulated by the "vuu-logging-level" cookie that is initialized in the worker.
Boilerplate conditional statements have been added around logs in the vuu-data package


vuu-ui/packages/vuu-data/src/server-proxy/array-backed-moving-window.ts
```diff
@@ -41,6 +45,12 @@ export class ArrayBackedMovingWindow {
   }
 
   setRowCount = (rowCount: number) => {
+    if (
```
heswell, line 48: there are a couple of issues here. Firstly, why incur the cost of a function call when the function is just returning a static value - just reference   loggingSettings.loggingLevel directly. Worse, we make 2 function calls.

Supposing you define the loglevels like this:
```
const LogLevel = {
 low: 1,
medium: 2,
high: 3
}
```

then , instead of having to write code like
```
if (logLevel === 'high' || logLevel === 'medium')
```

you can just use 
``
if (logLevel >= logLevel.medium)
``

vuu-ui/packages/vuu-data/src/server-proxy/buffer-range.ts
```diff
@@ -1,4 +1,9 @@
 import { VuuRange } from "@finos/vuu-protocol-types";
+import { logger } from "@finos/vuu-utils";
+
+const loggingLevel = () => {
```
heswell, line 4: this function is repeated in every file. I don't think we need it at all (see comment above) but if we wanted to keep it, it should be declared once and imported where needed.

vuu-ui/packages/vuu-utils/src/logging-utils.ts
```diff
@@ -9,6 +11,10 @@ export interface TableLogFn {
 	(properties?: object): void
 }
 
+type loggingSettings = {
+	loggingLevel: string;
+}
+
 export interface Logger {
```
heswell, line 18: this should be called a ConsoleLogger, because that's what it does,

heswell (changes requested):
I'm not sure that the 'high' , ''medium' categories add a lot of value. 
I think  these  represent what most logging systems achieve with

logger.log
logger.info
logger.debug
 
If we were to use that approach then we wouldn't need the checks around the log statements (which are actually function calls anyway), 

We could pass loggingLevel: 'info' for example, via the settings, then `logger.debug` statements would be noops, `logger.log` and `logger.info` would log.

vuu-ui/packages/vuu-utils/src/logging-utils.ts
```diff
@@ -9,68 +11,111 @@ export interface TableLogFn {
...
-// eslint-disable-next-line @typescript-eslint/no-empty-function
-const NO_OP: LogFn = () => {};
+
+const NO_OP = () => undefined;
+
+export const loggerNew = (category:string, logLevel: string) => {
+	const errorEnabled = logLevel === 'error';
+	const warnEnabled = errorEnabled || logLevel === 'warn';
+	const infoEnabled = errorEnabled || warnEnabled || logLevel === 'info';
+	const debugEnabled = errorEnabled || warnEnabled || infoEnabled || logLevel === 'debug';
```
heswell, line 35: the logic here is wrong, see the pseudocode in the linked issue.
For example, warning should not be enabled just because error is.
The logging levels form a hierarchy. If a particular level is set , all levels above are set.
"debug" is always the lowest level, with "debug" set you get ALL the log levels. If "info" is set, you get all logging except debug messages etc.

vuu-ui/packages/vuu-data/src/server-proxy/server-proxy.ts
```diff
@@ -795,8 +813,8 @@ export class ServerProxy {
             const viewport = this.viewports.get(clientViewportId);
             if (viewport) {
               viewport.setTableMeta(body.columns, body.dataTypes);
-            } else {
-              logger.warn(
+            } else if (log.warnEnabled) {
```
heswell, line 816: not sure if you've fully grokked the purpose of the code guards, you don't need to use them when logging a simple string.

https://commons.apache.org/proper/commons-logging/guide.html#Code_Guards

vuu-ui/packages/vuu-utils/src/logging-utils.ts
```diff
@@ -9,68 +11,111 @@ export interface TableLogFn {
...
 
+export type LogLevel = 'debug' | 'info' | 'warn' | 'error';
+
 export type BuildEnv = 'production' | 'development';
-// eslint-disable-next-line @typescript-eslint/no-empty-function
-const NO_OP: LogFn = () => {};
+
+const NO_OP = () => undefined;
+
+export const loggerNew = (category:string, logLevel: string) => {
```
heswell, line 31: I don't think the logLevel will vary across the app ? Thus no point baking it into each call to logger - might as well read it once inside the logging-utils module

heswell:
```
export const getCookieValue = (name: string): string | number | undefined => {
  if (global.document?.cookie !== undefined) {
    return global.document.cookie
      .split("; ")
      .find((row) => row.startsWith(`${name}=`))
      ?.split("=")[1];
  }
};

```

heswell (approved):

PR #579, opened 2023-03-21
UI filter panel - first pass

Joe-Dunleavy:
This is a draft PR for v1 of the UI filter panel.

**Features:**

1. **Filter types:**
- Typeahead filter for string columns. (Exact match or Starts with). The first option that is displayed in the list of suggestions is the starts with filter -- represented by whatever the user has typed followed by '...' (see pic below)
- Range filter for number columns

2. **Multiselect filters:** 
- Apply any amount of exact-match filters per column
- Up to one starts with filter per column.

3. **Highlight columns that are being filtered:** In the column selection dropdown, columns that have filter(s) applied are highlighted in blue
4. **Reset filters:** A "Clear" button that resets all filters

Note that the filter panel currently uses HTML elements and custom Icons instead of the existing salt components. These should be converted to salt components for a more unified look.

Starts with filter

### DIFF
--- a/vuu-ui/packages/vuu-filters/src/filter-panel/filter-panel.tsx
+++ b/vuu-ui/packages/vuu-filters/src/filter-panel/filter-panel.tsx
@@ -13,7 +13,7 @@ export const FilterPanel = ({ table, columns, onFilterSubmit }: Props) => {
   const [queries, setQueries] = useState<Query>({});
 
   useEffect(() => {
-    const queryString = getQueryString(queries);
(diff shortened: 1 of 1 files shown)


Joe-Dunleavy:
**UPDATE:** 
Since I created this draft PR, I've done some refactoring on this. 
I got rid of a lot of fluff and made the code more concise overall.
The TypeaheadFilter component could probably do with being broken down into smaller components as it is quite big, but I've left that with JJ.

PR #604, opened 2023-03-31
Issue 598 densities

theiceoan:
User is able to toggle between densities in Vuu
- the densities available are high, medium, low, and touch
A dropdown component is used to display and manipulate density


vuu-ui/packages/vuu-shell/src/density-provider/DensityProvider.tsx
```diff
@@ -0,0 +1,36 @@
+import { Dropdown, SelectionChangeHandler } from "@heswell/salt-lab";
```
heswell, line 1: Why is the file DensityProvider.tsx when the exported component is DensitySwitch ?
File should be DensitySwitch.tsx

vuu-ui/showcase/src/App.tsx
```diff
@@ -103,7 +105,8 @@ export const App = ({ stories }: AppProps) => {
             resizeable
             style={{ flexDirection: "column", flex: "1 1 auto" }}
           >
-            <Toolbar className="ShowcaseContentToolbar salt-density-high">
+            <Toolbar className={`ShowcaseContentToolbar salt-density-${density}`}>
```
heswell, line 108: I don't think its a good idea to set density on the Content Toolbar, we just want it on the Content area below. The toolbar is part of the ShowCase application chrome , it shouldn't jump about as user changes density. At some point we might want to offer the ability to set both theme and density on the entire Showcase app, but then we should offer an application level theme/density switcher on the very top toolbar. That can come later - we would have to make sure the Nav Tree supports all densitied

heswell (changes requested):
Few minor changes to make then we can merge. Also, need to rebase on latest main, as there are multiple changes around logging in vuu data

Issue #403, opened 2023-01-04
UI for Filters


heswell:
### Background
The Vuu server supports filtering via a text based filtering language, here are examples illustrating all the available operations

```sql
-- exact match against numeric or string type
lotSize = 100
currency = "EUR"

-- match against any one of a set, numeric or string type
currency in ["EUR", "GBP"]

-- and/or any number of sub clauses
currency = "EUR" and exchange = "XLON"
currency = "EUR" or exchange = "XLON"

-- GT, LT, GE, LE for numeric types
price > 1000
price >= 1000 and price <= 2000

-- Not Equal for string and numeric types
status != "closed"

-- partial matching for string types at start or end ( note no 'contains')
description starts "test"
name ends "-uk"

-- braces can be used to control precedence
currency = "EUR" and (exchange = "XLON" or status != "cancelled")  
```

At the VUU wire protocol level, filtering instructions can be applied to a viewport subscription at the point when the subscription is opened, i.e as part of the [CREATE_VP](https://github.com/finos/vuu/blob/main/vuu-ui/packages/vuu-protocol-types/index.d.ts#L250) message. A filter can be added, removed or updated on an existing subscription via the [CHANGE_VP](https://github.com/finos/vuu/blob/main/vuu-ui/packages/vuu-protocol-types/index.d.ts#L259) message. 

Within the UI code, interaction with the VUU data layer is via the [DataSource](https://github.com/finos/vuu/blob/main/vuu-ui/packages/vuu-data/src/remote-data-source.ts) object . This accepts `filter` as a property of the initial subscription options. It also offers a `filter` method to add, update and remove filtering on an open subscription. 

The data types associated with the columns of a viewport are relevant for filtering, some of the filter operators are not applicable to all column types. The types currently supported are:

- long
- int
- double
- string
- char

These are more Java than JavaScript friendly, JavaScript has no concept of long vs int for example. Note also that the long type can be used for timestamp values, so is typically used wherever dates are handled.
 
### Current support for filtering
Filter-related code is collected within `@finos/vuu-filters` and `@finos/vuu-filter-types`. The most commonly used types have been extracted into a separate package as they are referenced from multiple packages - any place an API can include a filter will use these types, so the datagrid and  data connectivity packages both require the `Filter` type, for example.

[...]

PR #613, opened 2023-04-04
Filter Panel UI 

jjgry:
This PR addresses issue #403 

### Features:

- Typeahead filter:
  - For string columns. 
  - The first option that is displayed in the list of suggestions is the starts with filter, represented by the user input followed by '...'
  - Up to one starts with filter per column.

image

- Range filter for number columns

image

- Multiselect filters:
  - Apply any amount of exact-match filters per column

image

 - Reset filters
   -  A "Clear" button that resets all filters


ColinEberhardt:
@jjgry I've added you to the Scott Logic organisation CLA

vuu/src/main/scala/org/finos/vuu/core/module/typeahead/TypeAheadRpcHandler.scala
```diff
@@ -39,7 +39,7 @@ class TypeAheadRpcHandlerImpl(val tableContainer: TableContainer) extends RpcHan
       case dataTable: DataTable =>
         dataTable.columnForName(column) match {
           case c: Column =>
-            dataTable.primaryKeys.foldLeft(Set[String]())(addUnique(dataTable, c, _, _)).filter(_.startsWith(starts)).toArray.sorted.take(10)
+            dataTable.primaryKeys.foldLeft(Set[String]())(addUnique(dataTable, c, _, _)).filter(_.toLowerCase().startsWith(starts.toLowerCase())).toArray.sorted.take(10)
```
jjgry, line 42: I think this probably needs to be removed as it's changing the core behaviour of filtering. Feedback would be appreciated

chrisjstevo: Hi @jjgry, 

Do you mean change the way casing is handled?  So you'd prefer to see it lower cased first before comparison?

Thanks,

Chris

jjgry: Hi Chris,

This is something that I added during development for my own convenience. I don't think this PR is the right place to change this, but it's worth considering the behavior. I think the expected behaviour is probably context-dependent and there is a general lack of filtering consistency in the software I've used. It's possibly worth considering options for the future of filtering in Vuu.

Best, JJ

heswell: I think we should either have case-insensitive filtering by default or add a flag which allow client to opt into case insensitive filtering. It's fine when a client picks values from a list, but when they are typing to filter that list down, or using starts or ends operators, I think its unfriendly to force them to know the right case

heswell: especially when fields like descriptions and titles might be inconsistent in their use of case  
in application data

jjgry: I've removed this for now so that this PR can progress and be merged. I agree, Steve, that for starts-with filters case insensitivity could be beneficial for lots of users. I'll leave this thought with you for now

PR #640, opened 2023-04-18
Feature context provider

theiceoan:
Salt Provider component has been replaced with a Vuu Theme Provider component that controls the available densities and theme modes for the showcase content.

### DIFF
--- a/vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
+++ b/vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
@@ -1,8 +1,7 @@
 import { Dropdown } from "@heswell/salt-lab";
 import { Density } from "@salt-ds/core";
-import { HTMLAttributes, useCallback, useContext } from "react";
+import { HTMLAttributes, useCallback } from "react";
 import cx from "classnames";
-import { ThemeContext } from "../theme-provider";
 
 const classBase = "vuuDensitySwitch";
 
@@ -13,16 +12,17 @@ export interface DensitySwitchProps
   extends Omit<HTMLAttributes<HTMLDivElement>, "onChange"> {
   defaultDensity?: Density;
   density?: Density;
+  onDensityChange: (density: Density) => void;
 }
 
 export const DensitySwitch = ({
   className: classNameProp,
   defaultDensity=DEFAULT_DENSITY,
+  onDensityChange,
 }:DensitySwitchProps) => {
-  const { setDensity } = useContext(ThemeContext)
   const handleSelectionChange = useCallback((_event, selectedItem) => {
-    setDensity(selectedItem);
-  }, [setDensity])
+    onDensityChange(selectedItem);
+  }, [onDensityChange])
 
   const className = cx(classBase, classNameProp);
 
--- a/vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
+++ b/vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
@@ -1,4 +1,4 @@
-import { Dropdown, SelectionChangeHandler } from "@heswell/salt-lab";
+import { Dropdown } from "@heswell/salt-lab";
 import { Density } from "@salt-ds/core";
 import { HTMLAttributes, useCallback, useContext } from "react";
 import cx from "classnames";
@@ -20,7 +20,7 @@ export const DensitySwitch = ({
   defaultDensity=DEFAULT_DENSITY,
 }:DensitySwitchProps) => {
   const { setDensity } = useContext(ThemeContext)
-  const handleSelectionChange:SelectionChangeHandler | string = useCallback((_event, selectedItem) => {
+  const handleSelectionChange = useCallback((_event, selectedItem) => {
     setDensity(selectedItem);
   }, [setDensity])
 
--- a/vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
+++ b/vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
@@ -1,7 +1,8 @@
 import { Dropdown, SelectionChangeHandler } from "@heswell/salt-lab";
 import { DEFAULT_DENSITY, Density } from "@salt-ds/core";
-import { HTMLAttributes, useCallback } from "react";
+import { HTMLAttributes, useCallback, useContext } from "react";
 import cx from "classnames";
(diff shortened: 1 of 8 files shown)


vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
```diff
@@ -1,28 +1,28 @@
...
+  density?: Density;
 }
 
 export const DensitySwitch = ({
   className: classNameProp,
   defaultDensity=DEFAULT_DENSITY,
-  onDensityChange
 }:DensitySwitchProps) => {
-
+  const { setDensity } = useContext(ThemeContext)
```
heswell, line 22: while this is fine, I think it makes for a nicer API to export a specialised hook (in the ThemeProvider) for consumers of density
```
const {setDensity} = useDensity()
```

So client do not need to reference the ThemeContext object at all.

vuu-ui/packages/vuu-shell/src/density-switch/DensitySwitch.tsx
```diff
@@ -1,28 +1,28 @@
...
   const handleSelectionChange:SelectionChangeHandler | string = useCallback((_event, selectedItem) => {
```
heswell, line 23: that `| string` inthe type looks a bit suspect

vuu-ui/packages/vuu-shell/src/theme-provider/ThemeProvider.tsx
```diff
@@ -2,41 +2,55 @@ import React, {
...
 import cx from "classnames";
 
-export const DEFAULT_DENSITY = "medium";
-export const DEFAULT_THEME = "salt-theme";
+export const DEFAULT_DENSITY:Density = "medium";
+export const DEFAULT_THEME:Theme = "salt-theme";
+export const DEFAULT_THEME_MODE:ThemeMode = "light";
 
 export type Density = "high" | "medium" | "low" | "touch";
+export type Theme = "salt-theme";
```
heswell, line 17: I think ultimately theme will need to be a string, as we will want to allow custom / third party themes

vuu-ui/packages/vuu-shell/src/theme-provider/ThemeProvider.tsx
```diff
@@ -48,27 +62,30 @@ const createThemedChildren = (
 };
 
 interface ThemeProviderProps {
-  children: ReactElement;
+  children?: ReactElement;
```
heswell, line 65: I think this should be required - it doesn't make sense to have an empty ThemeProvider

vuu-ui/packages/vuu-shell/src/theme-provider/ThemeProvider.tsx
```diff
@@ -48,27 +62,30 @@ const createThemedChildren = (
...
 }
 
 export const ThemeProvider = ({
   children,
-  density: densityProp,
-  theme: themeProp,
+  // theme: themeProp,
+  themeMode: themeModeProp,
+  density: densityProp
 }: ThemeProviderProps) => {
```
heswell, line 77: we've lost the code that inherits values from a higher level ThemeProvider. This is very important

vuu-ui/packages/vuu-shell/src/theme-provider/ThemeProvider.tsx
```diff
@@ -48,27 +62,30 @@ const createThemedChildren = (
...
-    useContext(ThemeContext);
+  const [themeMode, setThemeMode] = useState<ThemeMode>("light");
+  const [density, setDensity] = useState<Density>("high");
 
-  const density = densityProp ?? inheritedDensity ?? DEFAULT_DENSITY;
-  const theme = themeProp ?? inheritedTheme ?? DEFAULT_THEME;
-
-  const themedChildren = createThemedChildren(children, theme, density);
+  const currentDensity = densityProp ?? density ?? DEFAULT_DENSITY;
+  const currentThemeMode = themeModeProp ?? themeMode ?? DEFAULT_THEME_MODE;
```
heswell, line 82: if this was the correct behaviour, then a cleaner way to implement this would simple be like this
```
export const ThemeProvider({
   themeMode = DEFAULT_THEME_MODE
})
```
then just use `themeMode` in the call below. However, this isn't correct, so you can't do it that way. We're missing the (possible) inherited value, so it has to be done like the original code

vuu-ui/packages/vuu-shell/src/theme-switch/ThemeSwitch.tsx
```diff
@@ -26,9 +23,9 @@ export const ThemeSwitch = ({
   className: classNameProp,
   defaultMode: defaultModeProp,
```
heswell, line 24: I think, in the case of this component, we should assume controlled mode only. The ThemeContext is always going to be where this state is managed.

heswell (changes requested):

vuu-ui/packages/vuu-shell/src/theme-provider/ThemeProvider.tsx
```diff
@@ -51,24 +65,29 @@ interface ThemeProviderProps {
...
 export const ThemeProvider = ({
   children,
-  density: densityProp,
-  theme: themeProp,
+  // theme: themeProp,
+  themeMode: themeModeProp,
+  density: densityProp
 }: ThemeProviderProps) => {
-  const { theme: inheritedTheme, density: inheritedDensity } =
+  const { density: initialDensity, themeMode: initialThemeMode } =
```
heswell, line 78: we still have the same problem here as discussed previously. Both props and setState can be used by clients to control the state. It needs to be one or the other. 

Thinking this through more carefully, we cannot introduce useState into the provider, it must be managed via props, as in the original version. Inheritance via the nesting of ThemeProviders won't work properly if Providers manage state internally with useState.  

Lets discuss in person

vuu-ui/showcase/src/App.tsx
```diff
@@ -100,34 +100,39 @@ export const App = ({ stories }: AppProps) => {
             revealSelected
             source={source}
           />
-          <Flexbox
-            className="ShowcaseContentContainer"
-            resizeable
-            style={{ flexDirection: "column", flex: "1 1 auto" }}
+          <ThemeProvider
```
heswell, line 103: Having a ThemeProvider nested between a Flexbox and its child is breaking the layout - try resizing the panel on the left.
I need to make a change to the Layout components to allow ThemeProvider as a Layout warpper

PR #644, opened 2023-04-21
cli

harryhartley:
The most basic possible cli, everything removed outside of the ability to run the launchApp script


harryhartley:
After a lot of messing around trying to get something more complicated working, I've decided to take a new approach - I deleted the spiderweb of changes and merge conflicts and just rebuilt the most basic cli I possibly could. All this does is take in one argument as an example, then runs the launchApp script.

I've asked JJ to have a quick look at this PR too, though with most of the code being removed for simplicity there's not much to see right now.

Can be installed to run as 'vuu', or the new index.js file can be run with node.

heswell (commented):
Does the cli need to live at the top level of vuu-ui ? Could it all live, for example as a vuu-cli folder within tools, which would itself be an npm package ? 

Could you add a README with instructions for running the CLI, both as a dev who wants to test it out locally, and  for someone who wants to install the cli package globally. Obviously the latter will only be an option once we publish it.

PR #653, opened 2023-04-25
Connection metrics

theiceoan:
* Tracking the messages per second sent by the server 
* Displaying this in the showcase (subscribed to the prices table)


vuu-ui/packages/vuu-data/src/websocket-connection.ts
```diff
@@ -151,6 +153,7 @@ export class WebsocketConnection implements Connection<ClientToServerMessage> {
     const callback = this[connectionCallback];
     ws.onmessage = (evt) => {
       const vuuMessageFromServer = parseMessage(evt.data);
+      this.messages.push(evt.data);
```
heswell, line 156: rather than save the messages, can we simply increment a count.

Saving them incurs the cost of the array , the cost of pushing an object onto that array and prevents the messages being garbage collected until we reset the array. All we really care about is the count.

Issue #686, opened 2023-05-09
As a developer I'd like a standard interface for adding data to tables on the server


chrisjstevo:
## Feature Request

For editing data on the server, we'd like to use a similar mechanism as we use with menu actions, where the UI can use generic callback functions for setting data into particular fields. On order to do this we would need a workflow something like the below. 

```
Client                                                                                Server
------                                                                                ------
* Invoke Menu Action (ex. SetSeqNum). =>                    SetSeqNum on server rpc service gets called 
                                                                 <=                     Server returns an instruction to OpenAndEditViewPort(tablename, form)
* UI Sends CREATE_VP_REQUEST on named table  =>
(and renders into specified form)
* UI sends generic instructions to setdata                 =>
(ex. VIEWPORT_SET_CELL (key, column, value) 
each time a user modifies a value, 
this puts the data on the server
* When editing is complete, UI sends VP_SUBMIT_FORM  => this invokes a specific onFormSubmit call on the server side, the server then gets the data from the table and sends it further on. 
```
The key to this approach is that all the calls for the UI are generic, it doesn't have any extra definition for the calls.

heswell:
If we want to keep this entirely generic, and want a reasonable user experience, we need the ability to send a few additional pieces of metadata to the client.

- a title for the dialog that will be displayed
- some kind of explanatory text to present to the user so they understand what they are being asked to submit
- a user friendly label for each form field UI displays

ddimtirov:
Few more things we may want to specify for each field:

* Label (different from protocol field name)
* Help text (description)
* Enum choices or autocompletion list
* Optional/mandatory (separate from validation, as you want display indication too)
* Validation (regex?, standard patterns?, groups?) - alternatively, the validation may be a roundtrip to the server side

All except the last bullet are present in the form metadata for our PoC.

PR #687, opened 2023-05-10
#686 Added ability to edit cells and rows in grids or bind data to se…

chrisjstevo:
…ssion table.


vuu/src/test/scala/org/finos/vuu/viewport/sessiontable/EditSessionTableTest.scala
```diff
@@ -0,0 +1,151 @@
...
+      val baseTable = tableContainer.getTable(FIX_SEQ_RESET_TABLE)
+
+      val sessionTable = tableContainer.createSimpleSessionTable(baseTable, session)
+
+      val row = selection.rowKeyIndex.keys.map(selection.viewPort.table.pullRow(_)).toList.head
+
+      val processId = row.get("id").toString
+
+      sessionTable.processUpdate(processId, RowWithData(processId, Map("process-id" -> processId, "sequenceNumber" -> 0)), clock.now())
+
```
chrisjstevo, line 41: Should this RenderComponent be an url or something else...?

vuu/src/test/scala/org/finos/vuu/viewport/sessiontable/EditSessionTableTest.scala
```diff
@@ -0,0 +1,163 @@
...
+      OpenDialogViewPortAction(ViewPortTable(sessionTable.name, sessionTable.tableDef.getModule().name), RenderComponent.InlineForm)
+    }
+
+    override def menuItems(): ViewPortMenu = ViewPortMenu("Admin",
+      new SelectionViewPortMenuItem("Reset SeqNum", "", this.openEditSeqNum, "OPEN_EDIT_RESET_FIX")
```
heswell, line 46: this is defined as a Selection Menu Item. If user selects multiple rows before invoking action, will resetSequence be applied to all selected process Ids.
OR should this be a Row MenuItem ? Or is should we use Selection Menu Item and enforce single selection only in table ?

chrisjstevo: At the moment we take the head row, I think we should have a category of single selection only yes. I guess we can add that tho. At the moment we could send back a validation error if more than one row.

PR #727, opened 2023-05-31
Finos site styling

pranavsetty:
Override the current footer and add new a swizzled footer


website/src/theme/Footer/index.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import VuuLogo from "../../../static/img/finos/vuu-text.png"
+import FinosLogo from "../../../static/img/finos/finos.png"
+
+import "../../css/custom.css"
```
heswell, line 5: no need to import this. It's imported by the page that hosts this component.

heswell (changes requested):
- overall height of footer is 448px, it should be 330px
- - missing border
- missing 'copyright'
- vertical alignment of columns is out, they should be centred, not aligned top
- right padding is incorrect, looks like padding intended for container right has instead been applied to container top
- incorrect font used. should be 'Raleway'
- font color isn't right, should be 80%. Use --vuu-dark-text-secondary from the vuu-theme
- separator spacings are out, they need to be mid-way between columns
- all column alignments are out, partly down to the incorrect right padding, also because of not settings widths per the designs, see below (design above, implementation below)
Screenshot 2023-06-01 at 12 32 28

PR #800, opened 2023-06-28
Update cve-scanning.yml to run with dependency-check

josspo:
Changed Dependency Check jobs. 
- It will fail on any CVE higher or equal to 7
- It will upload a report in html


chrisjstevo:
Hi @josspo can I check where the report files end up?

Thanks.

PR #815, opened 2023-07-13
Demo html auth

heswell:
There are two entry points to the demo app

- login.html
- demo.html

Both entry points route user to index.html , once user is known (user is used to load correct UI)

login.html prompts for password as well as username. This is used to authenticate against Vuu `/api/authn` service. The demo Vuu server does not actually provide any authentication, it simply returns a token .

demo.html only prompt for username, used to load correct UI. 

a cookie is used to track which entry point was used, so `logout` returns user to appropriate entrypoint page.

default entrypoint , if user tried to access index.html directly without first identifying , is demo.html


ddimtirov (commented):
👍

Issue #778, opened 2023-06-20
Create Filter Clause Component


hyhydev:
## Feature Request
Filter Clause

### Acceptance Criteria:
* Column selection
* Operator selection?
* Numeric and text input
* Typeahead suggestion for text field

PR #819, opened 2023-07-17
Filter Clause Component

jjgry:
Filter clause component to satisfy issue https://github.com/finos/vuu/issues/778

image

Known issues:
* Dropdown list items are too tall. The default height is difficult to change while maintaining proper scroll behaviour. The desired height is 20px, currently they are the default at 36px
* The inputs are too wide, they should be fitting their content (the search text). This is difficult to to in plain CSS, so I'm seeking feedback as to how this can be achieved (likely by measuring some hidden elements with JS)
* `startsWith` and `endsWith` filters not supported yet


jjgry:
This is still a work in progress, I'm merging since it's unused and I won't be able to continue working on it.

PR #836, opened 2023-08-02
added FastBuildBranchesOfTreeOfRows - 100000 for first batch of rows for branch build

keikeicheung:


vuu/src/main/scala/org/finos/vuu/core/VuuServer.scala
```diff
@@ -106,6 +106,16 @@ class VuuServer(config: VuuServerConfig)(implicit lifecycle: LifecycleContainer,
     tableContainer.createAutoSubscribeTable(tableDef)
   }
 
+  def createBasketTable(tableDef: TableDef): DataTable = {
```
chrisjstevo, line 109: Hi Emily,

We don't need these explicit calls on the Vuu server, the createTable call turns a table def into a map backed real table. You can use the existing ones for your new tables.

keikeicheung: yes thanks! i have changed it

keikeicheung:
separating two tasks
- put the fast build tree branch to a branch @chrisjstevo - want to check whether that can be merged 
- basket tables: will be checking in from another pull request

PR #849, opened 2023-08-14
Feat/client side data

JoachimStanislaus:
- Client side functionalities ( Ability to work in tandem )
   - Filter
   - Sort
   - Groupby
   - Aggregate
      - Sum, High, Low, Average, Count, Distinct
      
- Persistent Aggregated Data Display.
- Dynamic Fake Data Generator


toolbox/test-graphviz.txt
```diff
@@ -0,0 +1,31 @@
+digraph regexp {
```
heswell, line 1: can you remove this file from the commit. It's an artifact of the server build. Should be in the gotignore

vuu-ui/package.json
```diff
@@ -68,7 +68,7 @@
     "eslint-config-prettier": "8.7.0",
     "eslint-plugin-cypress": "^2.13.3",
     "eslint-plugin-prettier": "4.2.1",
-    "eslint-plugin-react": "7.32.2",
```
heswell, line 71: why this change ?

vuu-ui/packages/vuu-data/src/array-data-source/array-data-source.ts
```diff
@@ -40,6 +44,8 @@ import {
 } from "../vuuUIMessageTypes";
 import { collapseGroup, expandGroup, GroupMap, groupRows } from "./group-utils";
 import { sortRows } from "./sort-utils";
+import { el } from "@faker-js/faker";
```
heswell, line 47: don't think we are using faker in here, are we ?

JoachimStanislaus: sorry about it! Forgot to change the tag to WIP -- shifting faker to another file to make it more modular

vuu-ui/packages/vuu-data/src/array-data-source/array-data-source.ts
```diff
@@ -1,6 +1,6 @@
 import { DataSourceFilter, DataSourceRow } from "@finos/vuu-data-types";
```
heswell, line 1: doesn't stop us merging, but we're still lacking a `set config` that will allow us to set multiple attributes at same time. Will be needed for full compatability with RemoteDataSource

JoachimStanislaus: Will quickly add it in in the coming week! I shifted off that after discussion with Steve that it would be more modular with the current setup.

vuu-ui/package.json
JoachimStanislaus: Think it was an auto dependency update, but I'll switch it back manually next push!

PR #976, opened 2023-11-17
#975 Add basket constituents as instrument to be added in Basket Trading

keikeicheung:


example/order/src/main/scala/org/finos/vuu/provider/simulation/SimulatedBigInstrumentsProvider.scala
keikeicheung, line 1: @chrisjstevo I think we should put DefaultProvider right?

example/basket/src/main/scala/org/finos/vuu/core/module/basket/BasketModule.scala
```diff
@@ -112,6 +115,22 @@ object BasketModule extends DefaultModule {
...
+      .addTable(
+        TableDef(
+          name = "basketInstruments",
+          keyField = "ric",
+          columns = Columns.fromNames("ric".string(), "description".string(), "bbg".string(), "isin".string(),
+            "currency".string(), "exchange".string(), "lotSize".int()),
+          VisualLinks(),
+          joinFields = "ric"
+        ),
+        (table, vs) => new CompositeProvider(new SimulatedBigInstrumentsProvider(table), new BasketInstrumentProvider(table)),
```
keikeicheung, line 127: like this? @chrisjstevo

PR #997, opened 2023-11-22
Issue 984 ignite cache initial sketch

rumakt:


rumakt:
To start the entire env, including standalone ingnite ( as stated in : https://github.com/finos/vuu/pull/997/files#diff-73acb7f268b26e6aaf52a54cd1a1650de8b2bf27716d8aa701cee43d354753ccR10-R12 )  :
1. Start the CacheNodeApplication - it starts ignite node - bear in mind you can create many instances of this service.
2. Start IgniteOrderLoader - it loads random parent/child orders into a cache. 
3. IgniteCacheQueryApplication - is just a basic app that connects to the cache and does some query - can be useful for testing.

PR #1018, opened 2023-11-26
Draft - Issue 989 big data support scaling vuu how to abstract vuu interfaces to allow different implementations of key classes by adding features on construction

chrisjstevo:
First example of a plugin structure for project.

### DIFF
--- a/feature/ignite-feature/pom.xml
+++ b/feature/ignite-feature/pom.xml
@@ -5,14 +5,14 @@
     <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.finos.vuu</groupId>
-        <artifactId>example</artifactId>
+        <artifactId>feature</artifactId>
         <version>0.9.36-SNAPSHOT</version>
+        <!--relativePath>../pom.xml</relativePath-->
     </parent>
 
-    <artifactId>main-java</artifactId>
+    <artifactId>ignite-feature</artifactId>
 
     <dependencies>
-
         <dependency>
             <groupId>org.finos.vuu</groupId>
             <artifactId>vuu</artifactId>
@@ -30,12 +30,6 @@
 
         <dependency>
             <groupId>org.finos.vuu</groupId>
-            <artifactId>basket</artifactId>
-            <version>0.9.36-SNAPSHOT</version>
-        </dependency>
-
-        <dependency>
-            <groupId>org.finos.vuu</groupId>
             <artifactId>price</artifactId>
             <version>0.9.36-SNAPSHOT</version>
         </dependency>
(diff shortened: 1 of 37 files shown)


feature/ignite-feature/pom.xml
```diff
@@ -0,0 +1,273 @@
...
+
+        <dependency>
+            <groupId>org.finos.vuu</groupId>
+            <artifactId>price</artifactId>
+            <version>0.9.36-SNAPSHOT</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.finos.vuu</groupId>
+            <artifactId>order</artifactId>
```
rumakt, line 39: Do we want to have 'features' depend on particular 'use-cases' like 'order' or 'price' ?

rumakt:
I like the 'feature' and 'plugin' idea. This is very good 1st step towards making vuu a composable project

naleeha:
I think this is good for understanding what it could potentially look like with the plugin concept & bootstraping relevant features etc.
I'm not sure about the feature concepts and having factor for each at that level.
I wonder we can make the construction of main concepts a bit simpler by making it less flexible - e.g. you always create a VirtualViewPort from a VirtualDataTable which knows to take in VirtualSort function etc 
Ensure we resolve the right implementation at creation using dependency injection rather than each "feature" being away which plugin it belongs to. We may want to re-use the implementation in multiple plugin as well.

Would it make sense as next step to have a spike of implementing the Virtual/BigData vuu solution using the ignite & once we have the second example, try and come up with the exact interfaces and how we can wire them up correctly?

plugin/ignite-plugin/src/main/scala/org/finos/vuu/table/virtualized/RollingCache.scala
```diff
@@ -0,0 +1,18 @@
+package org.finos.vuu.table.virtualized
+
+import org.finos.vuu.core.table.RowData
+import org.finos.vuu.table.virtualized.cache.GuavaRollingRowDataCache
+
+trait RollingCache[KEY, VALUE] {
```
rumakt, line 6: Wondering if 'RollingCache' is the most fortunate name for this trait ? The contract does not suggest anything 'rolling' here ? i.e. it can but does it have to ?

rumakt:
Good step forward, lets keep the ball rolling.

plugin/ignite-plugin/src/main/scala/org/finos/vuu/table/virtualized/RollingCache.scala
chrisjstevo: Agreed, renamed from Rolling to Windowed

PR #1055, opened 2023-12-06
#983 Simulated prices provider to have random small walk logic - fixe…

keikeicheung:
…d a big jump at doWidenBidAndAsk (ric not in states)


chrisjstevo:
Looks great, thanks @keikeicheung :-)

Issue #483, opened 2023-02-01
create custom CellRenderer for boolean columns


heswell:
default implementation renders boolean values as simple strings "true", "false"

create a graphical cell renderer that renders a checkbox.

PR #1057, opened 2023-12-07
#483 add CheckboxCell renderer for boolean columns

junaidzm13:
- also adds this renderer as the default for boolean columns in case no renderer is specified.


junaidzm13:
Screenshot 2023-12-07 at 2 08 07 PM

heswell (approved):

PR #1075, opened 2023-12-12
Add support for additional date/time formats

junaidzm13:


vuu-ui/packages/vuu-utils/src/date/helpers.ts
```diff
@@ -0,0 +1,22 @@
+import { DateTimeColumnTypeSimple } from "packages/vuu-table-types";
```
heswell, line 1: watch out for this `packages` should be @finos

This will have been auto inserted by vs code. Happens to me, though didn't used to. I've tried to fix it by creating tsconfig files in each package. Doesn't seem to have worked. The solution is correct configutation of typescript, so it correctly resolves workspace package references . Haven't got it right yet

heswell: (It doesn't cause a runtime problem if the import is a type only file, though it would if the file was ts)

heswell (approved):
nice work. Just that one auto-inserted import declaration to fix.

vuu-ui/packages/vuu-utils/src/date/helpers.ts
junaidzm13: Thanks for catching that! I can look into helping with the `tsconfig`

PR #1087, opened 2023-12-27
1073 enable date/time pattern selection through ColumnSettings

junaidzm13:
- converges date and time column types to one unified date/time type, consistent
  with how some major programming languages handles datetime.
- date/time pattern selection is only available for columns with "date/time" type.
- changes DateTimePattern to be an object instead of strings to enable
  simultaneous selection of both date and time patterns.
- also moves DateTimeColumnDescriptor to vuu-table-types package for consistency.

### DIFF
--- a/vuu-ui/packages/vuu-table-extras/src/column-formatting-settings/ColumnFormattingPanel.tsx
+++ b/vuu-ui/packages/vuu-table-extras/src/column-formatting-settings/ColumnFormattingPanel.tsx
@@ -1,7 +1,6 @@
 import {
   ColumnDescriptor,
   ColumnDescriptorCustomRenderer,
-  ColumnTypeFormatting,
   ColumnTypeRendering,
 } from "@finos/vuu-table-types";
 import { Dropdown, SingleSelectionHandler } from "@finos/vuu-ui-controls";
@@ -11,23 +10,22 @@ import {
   getCellRendererOptions,
   getConfigurationEditor,
   isColumnTypeRenderer,
-  isDateTimeColumn,
   isTypeDescriptor,
 } from "@finos/vuu-utils";
 import { FormField, FormFieldLabel } from "@salt-ds/core";
 import cx from "clsx";
 import { HTMLAttributes, useCallback, useMemo } from "react";
-import { NumericFormattingSettings } from "./NumericFormattingSettings";
-import { DateTimeFormattingSettings } from "./DateTimeFormattingSettings";
+import { BaseNumericFormattingSettings } from "./BaseNumericFormattingSettings";
+import { LongTypeFormattingSettings } from "./LongTypeFormattingSettings";
 import { FormattingSettingsProps } from "./types";
 
 const classBase = "vuuColumnFormattingPanel";
 
 export interface ColumnFormattingPanelProps
-  extends HTMLAttributes<HTMLDivElement> {
+  extends HTMLAttributes<HTMLDivElement>,
+    FormattingSettingsProps {
   availableRenderers: CellRendererDescriptor[];
   column: ColumnDescriptor;
-  onChangeFormatting: (formatting: ColumnTypeFormatting) => void;
   onChangeRendering: (renderProps: ColumnTypeRendering) => void;
 }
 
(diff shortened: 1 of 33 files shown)


junaidzm13:
Here's how the UI looks:

Screenshot 2023-12-27 at 8 02 08 PM

Screenshot 2023-12-27 at 8 02 43 PM

vuu-ui/packages/vuu-utils/src/date/types.ts
```diff
@@ -21,6 +32,8 @@ const isTimePattern = (pattern: string): pattern is TimePattern =>
   supportedTimePatterns.includes(pattern as TimePattern);
 
 export const isDateTimePattern = (
-  pattern?: string
+  pattern?: ColumnTypeFormatting["pattern"]
```
heswell, line 35: All good, just feels a bit verbose to me.
How do you feel about just letting the nulls/undefined flow through

```
const isDatePattern = (pattern?: string): pattern is DatePattern =>
  supportedDatePatterns.includes(pattern as DatePattern);

const isTimePattern = (pattern?: string): pattern is TimePattern =>
  supportedTimePatterns.includes(pattern as TimePattern);

export const isDateTimePattern = (
  pattern?: ColumnTypeFormatting["pattern"]
): pattern is DateTimePattern =>
  isDatePattern(pattern?.date) || isTimePattern(pattern?.time);
```

vuu-ui/packages/vuu-table-extras/src/column-formatting-settings/ColumnFormattingPanel.tsx
```diff
@@ -112,7 +104,7 @@ export const ColumnFormattingPanel = ({
       <div
         className={cx(classBase, className, `${classBase}-${serverDataType}`)}
       >
-        {contentForType}
+        {formattingSettingsForType}
```
heswell, line 107: yup, better name

vuu-ui/packages/vuu-table-extras/src/column-formatting-settings/ColumnFormattingPanel.tsx
```diff
@@ -123,3 +115,20 @@ export const ColumnFormattingPanel = ({
     </div>
   );
 };
+
+function formattingSettingsByColType(props: FormattingSettingsProps) {
+  const { column } = props;
+
+  if (isDateTimeColumn(column)) {
```
heswell, line 122: somewhere we ought to warn the user if they do something invalid, like assigning a date/time type to a column with a serverDatatype that is not `long`.  This is part of the weird dissonance arising out of the fact that we're mixing server assigned data types and client side config.

Maybe we ignore for now and assume we'll fix all this when we overhaul the metadata sent by server.

vuu-ui/packages/vuu-table-extras/src/column-formatting-settings/ColumnFormattingPanel.tsx
```diff
@@ -123,3 +115,20 @@ export const ColumnFormattingPanel = ({
...
+    return <DateTimeFormattingSettings {...props} column={column} />;
```
heswell, line 123: This is good but there's still a gap in functionality. We're showing the data/time config settings if the column has already been configured as a 
date/time type. But the config settings panel should be where we can actually assign date/time as the type to a column of type `long`

I see two possibilities

1) we populate the renderer dropdown with an alternative to the 'default long renderer', we add a 'default date/time' renderer. This would assign the date/time type to a `long` column.
technically we're not actually implementing a renderer, just a formatter.

2) We add an additional UI item, to allow toggle on a long column, so user can pick between `number` or `date/time` When thet switch to date/time, we render the new settings

wdyt ?

heswell (changes requested):
all looks good. Need to add the ability to switch a long to data/time in the settings panel.
I'd suggest adding a few date and time columns to `instruments-extended` in the test data package) Thats where we added stuff to test the boolean columns  - I extracted these to instruments-extended rather than leave them in instruments  when I realized later that the extra columns broke some existing stuff. Now that instruments-extended isd a separate table, we can add anything we want. 
If you view it in Tables/SIMUL/InstrumentsExtended in showcase, you can bring up the ColumnSettings -

vuu-ui/packages/vuu-table-extras/src/column-formatting-settings/ColumnFormattingPanel.tsx
junaidzm13: Yeah that's true. To me the second option sounds better (though the first one seems easier to implement, second might need passing down a separate action handler for handling type change).

Mixing up renderers with formatters could end up confusing devs, and it could also be harder for the users to navigate in case we have many renderers.

junaidzm13:
This is how the table/settings look:

heswell (approved):
looks good, plus tested it locally

PR #1095, opened 2024-01-03
#1096 add lint and typecheck to test-ui CI/CD workflow

junaidzm13:
##### add lint and typecheck to test-ui CI/CD workflow

- Our current CI/CD workflow is unable to catch any linting or typecheck errors. Adding these checks will help catch them before they are committed into main.
- This change also defines a new action `setup-vuu-ui` which abstracts away common steps for ui jobs leading to better modularity and reusability.

##### fix eslint related errors
- eslint will ignore inlined-worker.js since its auto-generated.
- lint script to also include .ts,.tsx files.
- for now I've changed `@typescript-eslint/no-unused-vars` to only emit a warning as there are quite a few places where this rule is violated. Fixing all those with this MR would result in quite a big change.
- we can do a separate follow-up change to fix all those and then revert to emitting errors for unused vars.

##### fix path casing for Tooltip.examples.tsx


heswell (approved):
some of the bits in here (like the commented out sections in chrome-devtools piece) were left in deliberately, but I think lets get rid of all the junk, it's all in git if we ever need it.

Nice to start the new year with a. cleanup

Issue #1074, opened 2023-12-11
Add Date filtering capability to FIlters


heswell:
This is blocked pending completion of the Filterbar and FilterClauseEditor components. Should be ready, with full set of tests,  in about one week.
Details to follow

heswell:
More work has been done on filters, not uyet 100% but they are now much more usable. A first set of cypress tests is now in place as well. To be much expanded, but its a start.

The usage of Salt components has been refactored so that we now only consume components from the salt `core` library , none from the salt `labs` library. `core` are the components that are considered to be production ready.

This leaves us with a dilemma wrt to a datpicker/data range picker, both of which I imagine we will need for comprehensive date handling in filters. THere is no data picker yet in Salst. ( there will be, also a data range picker, but not for some time).

There is a feature complete calendar control, in `labs`

I would suggest we fork the calendar control, add it to vuu-ui-controls. Use DropdownBase to build our own date picker and date range picker. Then we can either

1) adopt salt verisons whern these are available in `core`
2) contribute our components back to salt.

PR #1106, opened 2024-01-08
#1074 add DatePicker & DateRangePicker components (first-cut)

junaidzm13:
Summary of changes:
- fork /calendar dir from salt-ds to vuu-ui-controls
- make forked calendar component work with vuu
  - fixes imports that were no longer valid
  - adds missing `salt-ds/icons` package to vuu-ui-controls
  - fixes linting errors
  - fixes an alignment issue with the calendar
  - fixes issues with css module imports -> I've reverted to importing css directly and relying on classes for the styles to take effect. Same as what we do elsewhere in VUU.
- add DatePicker and DateRangePicker components

### DIFF
--- a/vuu-ui/packages/vuu-ui-controls/package.json
+++ b/vuu-ui/packages/vuu-ui-controls/package.json
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@salt-ds/core": "1.13.2",
+    "@salt-ds/icons": "1.9.1",
     "@finos/vuu-data-types": "0.0.26",
     "@finos/vuu-table-types": "0.0.26",
     "@finos/vuu-layout": "0.0.26",
--- a/vuu-ui/packages/vuu-ui-controls/src/calendar/internal/CalendarNavigation.tsx
+++ b/vuu-ui/packages/vuu-ui-controls/src/calendar/internal/CalendarNavigation.tsx
@@ -1,0 +1,293 @@
+import {
+  Button,
+  ButtonProps,
+  makePrefixer,
+  Tooltip,
+  useId,
+} from "@salt-ds/core";
+import { ChevronLeftIcon, ChevronRightIcon } from "@salt-ds/icons";
+import { clsx } from "clsx";
+import {
+  ComponentPropsWithRef,
+  forwardRef,
+  MouseEventHandler,
+  SyntheticEvent,
+} from "react";
+import { Dropdown, DropdownProps } from "../../dropdown";
+import { ListItem, ListItemType } from "../../list";
+
+import { useCalendarContext } from "./CalendarContext";
+
+import calendarNavigationCss from "./CalendarNavigation.css";
+import { DateValue, isSameMonth, isSameYear } from "@internationalized/date";
+import { formatDate, monthDiff, monthsForLocale } from "./utils";
+import { SelectionChangeHandler } from "../../common-hooks";
+import { useWindow } from "@salt-ds/window";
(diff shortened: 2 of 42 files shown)


junaidzm13:
How the components look:

Screenshot 2024-01-08 at 7 57 05 PM
Screenshot 2024-01-08 at 7 56 41 PM

vuu-ui/packages/vuu-ui-controls/src/date-picker/DatePicker.css
```diff
@@ -0,0 +1,23 @@
+.vuuPortal {
+  z-index: calc(var(--salt-zIndex-flyover) + 1);
```
junaidzm13, line 2: Should we use popper's z-index for salt-ds tooltips?

salt-ds tooltip uses `--saltTooltip-zIndex` (which is not specified in vuu) or otherwise falls back to `--salt-zIndex-flyover`.

Was thinking if instead we should set `--saltTooltip-zIndex` to vuu popper's z-index? Wdyt?

vuu-ui/packages/vuu-ui-controls/package.json
```diff
@@ -11,6 +11,7 @@
   },
   "dependencies": {
     "@salt-ds/core": "1.13.2",
+    "@salt-ds/icons": "1.9.1",
```
heswell, line 14: I don't think we need to specify this as a dependency. It's already a dependency of salt core.
We never want to user these directly ourselves. 
In fact, these icons are a bit of a headache, we don't really want to use Salt icons, we would rather use our own. Thats easy with the way we specify icons (in css) but not easy with salt componnets as they ship them as react components.

In our theme CSS you will finsd this

```
 .saltIcon {
    display: none;
} 
```

vuu-ui/packages/vuu-ui-controls/src/calendar/internal/CalendarNavigation.tsx
```diff
@@ -0,0 +1,284 @@
...
+  MouseEventHandler,
+  SyntheticEvent,
+} from "react";
+import { Dropdown, DropdownProps } from "../../dropdown";
+import { ListItem, ListItemType } from "../../list";
+
+import { useCalendarContext } from "./CalendarContext";
+
+import "./CalendarNavigation.css";
+import { DateValue, isSameMonth, isSameYear } from "@internationalized/date";
```
heswell, line 22: interesting. I assumed there would probably be a third party library for date manipulation. I didn't know this one , but I'm sure there was good reason for choosing it. I'll have to famikliarise myself with it

vuu-ui/packages/vuu-ui-controls/src/date-picker/types.ts
```diff
@@ -0,0 +1,25 @@
...
+  ZonedDateTime,
+} from "@internationalized/date";
+import { CalendarProps } from "../calendar/Calendar";
+import { RangeSelectionValueType } from "../calendar";
+
+export type PickerSelectionType = DateValue | RangeSelectionValueType;
+
+export interface BaseDatePickerProps<T = PickerSelectionType>
+  extends Pick<CalendarProps, "hideOutOfRangeDates" | "hideYearDropdown"> {
+  onSelectedDateChange: (selected: T) => void;
```
heswell, line 14: probably good to keep APIs as similar as possible across form controls, so this could be just onSelectionChange to match list, dropdown,  combobox etc

heswell (approved):

PR #1117, opened 2024-01-11
#1074 fix bugs in FilterBar

junaidzm13:
- fixes a bug that crashes the app when a user tries editing an existing filter. Cause: we try finding the edited filter in `filters state` by reference but the reference has already been changed since `editFilter` is itself a state and hence the reference changes on every set.
- adds resetting the `editingFilter.current` to `undefined` after apply save. Otherwise, crashes the app when the user has already successfully edited an existing filter and tries to add a new one.
- lastly, also fixes an issue with `TextInput`'s multiselection handling, where it was wrongly sending a single value instead of an array of length 1. Crashes the app when the user selects only a single element in a multi value filter clause.


heswell (approved):
nice work

PR #1123, opened 2024-01-12
#1074 fix bug when editing filters

junaidzm13:
- `onSelectionChange` on `ExpandoCombobox` only gets called when a selection changes, which cause issues when someone tries to select the already selected item from the list dropdown. With this change I've provided a `onListItemSelect` handler which is same as `onSelect` on `List`.
- Also, text selection used to "appear" blank in `ExpandoCombobox` because the background & text both used to change to white on selection. Changed the background color on selection to a shade of blue.

### DIFF
--- a/vuu-ui/packages/vuu-ui-controls/src/combo-box/ComboBox.tsx
+++ b/vuu-ui/packages/vuu-ui-controls/src/combo-box/ComboBox.tsx
@@ -15,7 +15,7 @@ import {
   useCollectionItems,
 } from "../common-hooks";
 import { DropdownBase, DropdownBaseProps } from "../dropdown";
-import { List, ListProps } from "../list";
+import { List, ListControlProps, ListProps } from "../list";
 import { ChevronIcon } from "../list/ChevronIcon";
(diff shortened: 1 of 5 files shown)


vuu-ui/packages/vuu-ui-controls/src/combo-box/ComboBox.tsx
```diff
@@ -207,7 +213,7 @@ export const ComboBox = forwardRef(function Combobox<
           highlightedIndex={highlightedIndex}
           itemTextHighlightPattern={String(inputProps.value) || undefined}
           listHandlers={listHandlers}
-          onSelectionChange={onSelectionChange}
+          onSelectionChange={onSelectionChange} // not really needed, since onClick in listHandlers will be used instead.
```
junaidzm13, line 216: @heswell let me know if I should get rid of this, I've checked it's not used if `listHandlers` is not undefined. And it is not undefined here.

heswell: yes, lets get rid of it

PR #1130, opened 2024-01-15
#1074 enable date filtering using DatePicker

junaidzm13:
#### 1) enable date filtering using DatePicker

- adds `tableConfig` props to `FilterBar` as column descriptors are
  required to differentiate between a date/time and a long column.
- since for now we're assuming that server side filter handling would
  stay unchanged (treated same as long), had to add some extra mapping
  where we convert filter object to query (filterAsQuery). It is to
  handle special cases of `=` and `!=` with dates as epoch-milis.
- This change also fixes some of the bugs with the way we displayed
  tooltip (with details of applied filters) on FilterPill. Now it
  also displays the heirarchy between multi clause filters which
  wasn't visible before.

#### 2) fix bugs related to editing multi-clause filters

Bugs fixed:
1) it errored when editing a multi clause filter -> because when focus
   changed to a non-first edited filter clause that changed active index
   to include those indices, but when we saved the change it tried
   accessing filter at that index it was undefined (multi clause
   filter is considered as one when not being edited but as multiple
  when edited).
2) now we also keep name unchanged when a filter is edited, since the user
   would find it annoying to rename filter everytime a user edits it.

#### 3) add FilterBar component tests with cypress

- fixes cy test files types/linting issues by removing `*.cy.*`
  from excluded files in `tsconfig.json`
- also fixes failing `Tabstrip.cy.tsx` tests. Now all cypress
  component tests are passing.


vuu-ui/packages/vuu-filters/src/filter-clause/DateInput.tsx
```diff
@@ -0,0 +1,62 @@
...
+      hideOutOfRangeDates
+    />
+  );
+};
+
+function getInitialState(value: Props["value"]) {
+  return value ? toCalendarDate(new Date(value)) : undefined;
+}
+
+const getEpochMillisConverter =
```
heswell, line 46: maybe put this in vuu-utils/date/utils ?

vuu-ui/packages/vuu-filters/src/filter-clause/useFilterClauseEditor.ts
```diff
@@ -160,24 +158,24 @@ export type FilterClauseCancelHandler = (
   reason: FilterClauseCancelType
 ) => void;
 export interface FilterClauseEditorHookProps {
+  columnDescriptors: Record<string, ColumnDescriptor>;
```
heswell, line 161: maybe use the named type you defined for this

vuu-ui/packages/vuu-filters/src/filter-clause/FilterClauseEditor.tsx
```diff
@@ -60,62 +59,13 @@ export const FilterClauseEditor = ({
...
-    InputProps,
-    filterClause,
-    onDeselectValue,
-    onChangeValue,
-    suggestionProvider,
-    table,
-    value,
-  ]);
+  const columns = useMemo(
+    () => Object.values(columnDescriptors),
```
heswell, line 66: One consequence of switching from tableSchema.columns to tableConfig.columns is that whereas previously we could add any column from the Vuu table to our filter, now we can only include columns actually present in our Table (which includes hidden columns).   
I think this would probably be the expected behaviour (so an improvement) , though its worth at least thinking about.

heswell (changes requested):
great work. Couple of places where I think variable name changes would help avoid confusion. The only real MUST FIX are the package references in import statement,s.

Feel free to merge soon as you push change. If you don't have permission, ping me on slack tomorrow and I'll do it.

vuu-ui/packages/vuu-filter-types/index.d.ts
```diff
@@ -37,10 +37,11 @@ export interface SingleValueFilterClause<T = string | number | boolean>
   value: T;
 }
 
-export interface MultiValueFilterClause extends NamedFilter {
+export interface MultiValueFilterClause<T = string[] | number[] | boolean[]>
```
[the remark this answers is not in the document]
junaidzm13: With that it considers `values` as an `Array<string | number | boolean>`, which isn't what we want. We might have to revisit the overall filter types later

vuu-ui/packages/vuu-filters/src/filter-bar/useFilterBar.ts
```diff
@@ -69,6 +72,11 @@ export const useFilterBar = ({
   >();
   const [promptProps, setPromptProps] = useState<PromptProps | null>(null);
 
+  const columnMap = useMemo(
```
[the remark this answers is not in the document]
junaidzm13: Changed it to `columnsByName: ColumnDescriptorsByName`, wdyt?

vuu-ui/packages/vuu-filters/src/filter-bar/useFilterBar.ts
```diff
@@ -452,6 +460,7 @@ export const useFilterBar = ({
     addButtonProps,
     editFilter,
     filters,
+    columnDescriptors: columnMap,
```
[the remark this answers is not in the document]
junaidzm13: I've changed it to `columnsByName: ColumnDescriptorsByName`

vuu-ui/packages/vuu-filters/src/filter-clause/useFilterClauseEditor.ts
junaidzm13: Good point. Didn't have any type defined before, but now I do! `ColumnDescriptorsByName`

vuu-ui/packages/vuu-filters/src/filter-bar/FilterBar.tsx
```diff
@@ -37,12 +39,14 @@ export const FilterBar = ({
   onFiltersChanged,
   showMenu: showMenuProp = false,
   tableSchema,
+  tableConfig,
```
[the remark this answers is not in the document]
junaidzm13: Agreed!

vuu-ui/packages/vuu-filters/src/filter-clause/DateInput.tsx
junaidzm13: I did think about it previously (either `date/utils` or `filter/utils`), but I eventually decided not to as I think that functionality is tightly coupled with filtering (specifically with the `DateInput` component itself) - it takes in an `op` and returns start of day for `=` & `!=` which isn't necessarily correct if viewed in isolation - so it won't be much of use elsewhere. Wdyt?

vuu-ui/packages/vuu-filters/src/filter-clause/FilterClauseEditor.tsx
junaidzm13: Ahh good to know this!

vuu-ui/packages/vuu-filters/src/filter-clause/DateInput.tsx
heswell: fair enough. Agreed.

heswell (approved):

PR #1131, opened 2024-01-15
make sure combobox respects defaultSelected

heswell:
I've introdued a suite of cypress tests (from Salt) for ComboBox. They are not all passing yet so I've left some commented out. I don' think any of the functionality that still needs attention is anything we urgently need (page up and page down in combobox, disabled combobox, readonly combobox etc)

I'll return to these to get all tests passing eventually, but for now there should be enough to unblock the Vuu Filters


vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
```diff
@@ -126,6 +126,22 @@ export const ExpandoCombobox = forwardRef(function ExpandoCombobox<
     [itemToString, onSelectionChange]
   );
 
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const getDefaultSelected = (): any => {
```
heswell, line 131: any until I can get the type magic right

vuu-ui/packages/vuu-filters/src/filter-clause/TextInput.tsx
```diff
@@ -52,6 +52,10 @@ export const TextInput = forwardRef(function TextInput(
   }: TextInputProps,
   forwardedRef: ForwardedRef<HTMLDivElement>
 ) {
+  console.log(`TextInput ${value} operator = ${operator}`);
```
junaidzm13, line 55: You probably forgot to remove this log?

vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
```diff
@@ -126,6 +126,22 @@ export const ExpandoCombobox = forwardRef(function ExpandoCombobox<
...
+    if (initialValue.current === undefined) {
+      return undefined;
+    } else if (Array.isArray(initialValue.current)) {
+      return props.source?.filter((item) =>
+        initialValue.current.includes(itemToString(item))
+      );
+    } else {
+      return props.source?.find(
+        (item) => itemToString(item) === initialValue.current
+      );
```
junaidzm13, line 141: I think `initialValue.current` will always be a string (as passed in `value` will always be a string). So we do not need the `Array.isArray` check.

But then we might not be handling the multi-select case. Would it be better if we get the `defaultSelected` from `ExpandoCombobox` as a prop? And then we can just create a ref in here and pass that to `Combobox`? That way we leave the decision to the consumer (who'd know whether it's a single/multi select)

vuu-ui/packages/vuu-ui-controls/src/combo-box/useCombobox.ts
```diff
@@ -274,7 +283,7 @@ export const useCombobox = <Item, S extends SelectionStrategy>({
   );
 
   const handleKeyboardNavigation = useCallback(() => {
-    setDisableAriaActiveDescendant(false);
+    // setDisableAriaActiveDescendant(false);
```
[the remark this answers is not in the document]
heswell: I'll remove this handler. tbh I was struggling to understand what the initial purpose of the code was. As far as I can see we only want to disable aria activeDescendant when combo is closed, which is simple. If we later find we need it under other circumstances, we can re-introduce.

vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
heswell: good spot, this is incorrect.

heswell: Looking more closely, the reason the code works is because at runtime the value passed in CAN actually be an array. It's the value attribute from the Filter clause, which is a `FilterClauseValue` which can be an array.

I'm confusing everything and invalidating the type checking with this (in FilterClauseEditor)
```
value={value as string}
```
Was probably meant to be a temp hack while I got things working but forgot about it.

I'll fix the types

vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
```diff
@@ -52,8 +54,8 @@ export const ExpandoCombobox = forwardRef(function ExpandoCombobox<
   const { itemToString = defaultToString } = props;
   const initialValue = useRef(value);
 
-  const itemsToString = useCallback(
-    (items: Item[]) => {
+  const itemsToString = useCallback<<I = Item>(items: I[]) => string>(
+    (items) => {
       const [first, ...rest] = items;
       if (rest.length) {
         return `${itemToString(first)} + ${rest.length}`;
```
junaidzm13, line 61: We might be able to do something like:

`const itemsToString = useCallback(
    (items: Array<Item> | Array<string>) => {
      const [first, ...rest] = items;
      const firstAsString =
        typeof first === "string" ? first : itemToString(first);
      if (rest.length) {
        return `${firstAsString} + ${rest.length}`;
      } else {
        return firstAsString;
      }
    },
    [itemToString]
  );`
  
  this way we can keep the type to `itemToString?: (i: Item) => string`, and similarly avoid type casting such as `column as ColumnDescriptor`
  
  What do you think? If you want we can do this in a separate follow-up change.

vuu-ui/packages/vuu-ui-controls/src/combo-box/useCombobox.ts
```diff
@@ -273,10 +281,6 @@ export const useCombobox = <Item, S extends SelectionStrategy>({
     [handleOpenChange, isMultiSelect]
   );
 
-  const handleKeyboardNavigation = useCallback(() => {
-    setDisableAriaActiveDescendant(false);
-  }, []);
-
```
junaidzm13, line 279: Always a treat to see code being removed 🏆

vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
heswell: I like it, lets do it after I merge this one.

Issue #1138, opened 2024-01-22
Add equal filtering support for columns of type `Long`


junaidzm13:
### Steps to Reproduce:
Apply an `=` or `!=` filter on a long column and it will throw an exception.

It's because we do not check for `Long` in `match` statement inside `EqualsClause case class`.

### Expected Result:
It should handle `=` and `!=` for columns of type `Long` and apply filters as expected.

PR #1139, opened 2024-01-22
#1138 support `Long` columns in EqualsClause filter method

junaidzm13:
- this change also adds tests for EqualsClause filter method to make sure that this bug doesn't get re-introduced.

### DIFF
--- a/vuu/src/test/scala/org/finos/vuu/core/filter/FilterClauseTest.scala
+++ b/vuu/src/test/scala/org/finos/vuu/core/filter/FilterClauseTest.scala
@@ -1,0 +1,81 @@
+package org.finos.vuu.core.filter
+
+import org.finos.vuu.core.table.RowWithData
+import org.scalatest.featurespec.AnyFeatureSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks._
+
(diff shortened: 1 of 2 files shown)


naleeha:
image
I think the test results looks really good, very descriptive & we get test for each case

vuu/src/test/scala/org/finos/vuu/core/filter/FilterClauseTest.scala
```diff
@@ -0,0 +1,60 @@
...
+      ("Int", "size", "7", false),
+      ("Long", "timestamp", "5", true),
+      ("Long", "timestamp", "0", false),
+      ("Float", "priceInFloat", "100.0", true),
+      ("Float", "priceInFloat", "100.1", false),
+      ("Double", "priceInDouble", "100.0", true),
+      ("Double", "priceInDouble", "100.1", false),
+      ("Boolean", "true-col", "true", true),
+      ("Boolean", "true-col", "false", false),
+      ("Null", "null-col", "null", false)
```
naleeha, line 37: I think there is two more edge cases here
1. if the there is a column with value null and we call filter(null) i would expect it to return true.
2. If there is a column that is no in the test data and you try to filter for that column i guess we need to report that as error.. not sure what it does currently 

The second one i would write its own test as its edge cases

naleeha (approved):

PR #1141, opened 2024-01-22
manage named filters in application settings

heswell:
I had previously added code to vuu-filters package to store named filters in application settings. This was the wrong place to do this. The FilterBar should not know or care about the fact that we might persist filters across sessions and it should not require a dependency on vuu-shell.

I have extracted this code to the FilterTable feature. The feature is responsible for loading/saving   persisted state data. It now handles filters as well as other options. It also populates the context menu location 'filter-bar-menu' with saved menu items appropriate to the current tableSchema.

Opens a new question - should the filterbarmenu (the icon at far left) even be part of filterbar, we could extract that to nthe Feature as well ?

Right now, we are not doing anything with the  list of named filters that can be dropped down from the filter menu. That will follow in a separate PR.

Need to decide how we present these named filters. for example

- what if they are already in the FilterBar, do we show them greyed out ?
- OR do we show them as a multiselect, where the filters inn the filterbar appear as selected ?

The code in vuu-shell pertaining to LayoutManagement should all be changed to refer to `Persistence` not `Layout`, this has begun but there are still some entities, like `useLayoutManager` that should be `usePersistenceSomethingOrOther` The code that actually pertains to Layout (like LayoutProvider) all lives in the vuu-layout package.

WIll submit seperate PR just for the renaming


vuu-ui/sample-apps/feature-filter-table/src/useFilterTable.tsx
```diff
@@ -59,6 +78,84 @@ export const useFilterTable = ({ tableSchema }: FilterTableFeatureProps) => {
     [dataSource]
   );
 
+  const hasFilterWithName = (filters: NamedFilter[], name: string) =>
+    filters.findIndex((f) => f.name === name) !== -1;
+
```
junaidzm13, line 83: Should we move this out of the hook as a helper function, since it doesn't really depend on any props or state?

vuu-ui/sample-apps/feature-filter-table/src/useFilterTable.tsx
```diff
@@ -39,8 +42,24 @@ type FilterTableConfig = {
...
 
+type SavedFilterMap = {
+  [key: string]: Omit<NamedFilter, "name"> & { name: string }[];
+};
+
 export const useFilterTable = ({ tableSchema }: FilterTableFeatureProps) => {
   const { dispatch, load, save } = useViewContext();
+  const { getApplicationSettings, saveApplicationSettings } =
+    useLayoutManager();
+  console.log({ getApplicationSettings, saveApplicationSettings });
```
junaidzm13, line 53: should we remove this `console.log`?

PR #1152, opened 2024-01-23
Issue 40 large dataset

heswell:
UI Table support for large datasets (> 1 million rows)

supports scrolling through full dataset with
- scrollbar
- scrolltrack
- scrollwheel
- page up/down
- end key/home key
- arrow key up/down


vuu-ui/packages/vuu-table/src/useTableScroll.ts
```diff
@@ -264,13 +330,14 @@ export const useTableScroll = ({
...
-                // const from = firstRowRef.current + 1;
-                // console.log(`setRange from ${from}`);
-                // setRange({ from, to: from + viewportRowCount + 1 });
+                const offset = direction === "down" ? 1 : -1;
+                onVerticalScrollInSitu?.(offset);
+                const firstRow = firstRowRef.current + offset;
+                firstRowRef.current = firstRow;
+                setRange({
+                  from: firstRow,
+                  to: firstRow + viewportRowCount + 1,
```
junaidzm13, line 339: In other places, we're not adjusting index by `+ 1` for `to:` when setting range. Is this intentional here?

vuu-ui/packages/vuu-utils/src/range-utils.ts
```diff
@@ -12,12 +12,21 @@ interface FromToRange {
...
+  r1: VuuRange | undefined,
+  r2: VuuRange | undefined
+) => {
+  return r1?.from === r2?.from && r1?.to === r2?.to;
+};
+
 export function getFullRange(
   { from, to }: VuuRange,
   bufferSize = 0,
   rowCount: number = Number.MAX_SAFE_INTEGER
```
junaidzm13, line 25: Does this represent total row count? If yes, should we rename it to `totalRowCount` for better readability? Wdyt?

junaidzm13 (approved):

vuu-ui/packages/vuu-utils/src/column-utils.ts
```diff
@@ -406,6 +406,28 @@ export const sortPinnedColumns = (
...
+      pinnedWidthLeft += visibleWidth;
+    } else if (pin === "right") {
+      pinnedWidthRight += visibleWidth;
+    } else {
+      unpinnedWidth += visibleWidth;
+    }
+  }
+  return {
+    pinnedWidthLeft: pinnedWidthLeft + 4,
+    pinnedWidthRight: pinnedWidthRight + 4,
```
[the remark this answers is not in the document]
heswell: indeed. I'll add a constant for now. This might need further work as the 4px selection `edge` is somewhet specific to the current design.

vuu-ui/packages/vuu-table/src/useTableScroll.ts
heswell: This used to be the norm, but I realised when writing tests that the +1 was not correct. This is a leftover.

vuu-ui/packages/vuu-utils/src/range-utils.ts
heswell: agreed. Have renamed it. As you will have seen, rowCount sometimes refers to number of rows in viewport. Good to be explicit.

PR #1156, opened 2024-01-24
#1074 use DateFormatter from @internationalized/date 

junaidzm13:


vuu-ui/packages/vuu-utils/src/date/formatter.ts
```diff
@@ -1,8 +1,9 @@
+import { DateFormatter } from "@internationalized/date";
```
heswell, line 1: Hesitated over this, just because it takes vuu-utils package from a zero dependency to a with-dependencies package.
It does make sense though to keep date formatting utils in here, so we can use them from cell rendrers etc.

heswell (approved):

PR #1160, opened 2024-01-25
#1074 refactor & add state change listeners to FilterBar 

junaidzm13:


junaidzm13:
@heswell when merging, can we please not squash these since the commits are independent. Thanks!!!

heswell (approved):
lgtm

Issue #989, opened 2023-11-21
Big Data Support - Scaling Vuu - How to abstract Vuu interfaces to allow different implementations of key classes by adding "features" on construction.


chrisjstevo:
### Discussed in https://github.com/finos/vuu/discussions/971

<div type='discussions-op-text'>

<sup>Originally posted by **chrisjstevo** November 15, 2023</sup>
### Introduction

Vuu is currently a largely in memory, near caching solution for UIs. In order to allow Vuu to scale to large data sets (10bn rows) we cannot cache all the data in memory, we need to be able to use big data stores, and move a remote cursor around in that data set as users scroll. 

In order to facilitate this requirement we have started to think about Vuu rather than as a deeply entwined 'core' library, but more as a series of internal interfaces. Some examples of these interfaces are: 

```
DataTable
ViewPort
ViewPortKeys() //work in progress
Provider
JoinManager
LoginValidator
Authenticator
```

If we take the first 5 interfaces, these relate to how we store and join data in memory. When we move to a paradigm where all the data is not in memory, but simply a window of the data is in memory, we will need to provide different implementations of these interfaces to facilitate this. 

### Features & Plugins

Features, as described here, are a way of grouping alternative implementations to some or all of the above interfaces with non standard implementations. For example, we may want to have the situation where we want to back Tables and Viewports by a Big Data store. In this case we would want to have some construct, conceptually like below, where a user could "plugin" a different implementation to use. 

Example: 

```scala
  //in a normal config object
  val config = VuuServerConfig(
    VuuHttp2ServerOptions()
      //only specify webroot if we want to load the source locally, we'll load it from the jar
     //</snip>
    VuuSecurityOptions()
      .withAuthenticator(authenticator)
      .withLoginValidator(new AlwaysHappyLoginValidator),
    VuuThreadingOptions()
      .withViewPortThreads(4)
      .withTreeThreads(4)
  ).withModule(PriceModule())
    .withModule(BasketModule())
    //add big data plugin 
    withPlugin(BigDataTableAndViewportPlugin())
```
We would then have a plugin type, that would allow the provision of some or all of these interfaces: 

```scala
class BigDataTableAndViewportPlugin{
     
         registerFeature[DataTable](BigDataTableFeature())
         registerFeature[JoinManager](BigDataJoinManagerFeature())
}
```

 </div>

PR #1164, opened 2024-01-26
#989 add support of remaining ops for ignite filters

junaidzm13:
#989 extend Ignite SQL filters to support other ops

- adds support for In, Starts, Ends, Not Equal, and Lesser Than
  operations.
- uses `(` & `)` for OR/AND clause to make sure that the filters
  are parsed and applied as intended and their order is preserved
  as (A OR B) AND C != A OR (B AND C).
- adds support for CharDataType columns where it made sense
  (i.e. using chars with Starts/Ends doesn't make sense)

### DIFF
--- a/plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/filter/IgniteSqlFilterClause.scala
+++ b/plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/filter/IgniteSqlFilterClause.scala
@@ -1,39 +1,91 @@
 package org.finos.vuu.feature.ignite.filter
 
+import com.typesafe.scalalogging.StrictLogging
 import org.finos.vuu.api.TableDef
-import org.finos.vuu.core.table.DataType.StringDataType
+import org.finos.vuu.core.table.DataType.{CharDataType, StringDataType}
+import org.finos.vuu.feature.ignite.filter.IgniteSqlFilterClause.EMPTY_SQL
+
+private object IgniteSqlFilterClause {
+  val EMPTY_SQL = ""
+}
 
 trait IgniteSqlFilterClause {
   def toSql(tableDef: TableDef): String
 }
 
-case class OrIgniteSqlFilterClause(clauses:List[IgniteSqlFilterClause]) extends IgniteSqlFilterClause{
+case class OrIgniteSqlFilterClause(clauses:List[IgniteSqlFilterClause]) extends IgniteSqlFilterClause {
   override def toSql(tableDef: TableDef): String = {
-    clauses.head.toSql(tableDef) //todo implement properly. just making it work as root clause for now
+    val sql = clauses.map(c => c.toSql(tableDef)).mkString(" OR ")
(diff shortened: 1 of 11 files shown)


plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/filter/IgniteSqlFilterClause.scala
```diff
@@ -1,39 +1,91 @@
...
+      case _ =>
+        logger.error(s"`Ends` clause unsupported for non string column: `$columnName`")
+        EMPTY_SQL
+    }
+}
+
+case class InIgniteSqlFilterClause(columnName: String, values: List[String]) extends IgniteSqlFilterClause with StrictLogging {
+  override def toSql(tableDef: TableDef): String = {
+    val processedValues = tableDef.columnForName(columnName).dataType match {
+      case CharDataType | StringDataType => values.map(quotedString.apply)
```
naleeha, line 82: Not sure what is considered better practice but if you want to call it quotedString like everywhere else we could do
values.map(quotedString(_))
Or
values.map(v=>quotedString(value))

plugin/ignite-plugin/src/test/scala/org/finos/vuu/feature/ignite/filter/IgniteSqlFilteringTest.scala
naleeha, line 1: this was my bad but Scenario("AND clause for two conditions") is a duplicate test now as we have Feature("Parse and apply AND filter")

naleeha (approved):
Looks good, good test coverage & like the code improvements

For AND and OR clause, would be good to cover case where one subcondition is supplied e.g. "ric = "VOD.L" and". This should not happen so FilterClause will treat it as unsupported use case.

One other case i am curious of is when user want to use statement like - condition A OR condition B AND condition C with no ( )
Technically then we want to translate to matching sql, SQL will know to execute the AND condition before the OR


Spoke with Steve and for the unsupported cases, it would be good to bubble up the errors to the user - i.e. return a rcp response with the error 

To do this we need to see whether there is easy way to return an errored results from these filter clauses that result in filter failing & returning appropriate rcp results. Might be worth creating a new issue ticket for this.

plugin/ignite-plugin/src/test/scala/org/finos/vuu/feature/ignite/filter/IgniteSqlFilteringTest.scala
junaidzm13: I've removed this, and added a scenario for support of `DOUBLE`.

plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/filter/IgniteSqlFilterClause.scala
junaidzm13: I like this `values.map(quotedString(_))` better, concise and neat. Updated ✅

PR #1174, opened 2024-01-30
fix filter clause editor bug when no suggestions returned

heswell:
fix bug encountered in FilterClauseEditor when typeahead service returns no suggestions
allow data to be typed and ENTER pressed to continue when typing text into combobox in filters
fix issues in list/combobox/dropdown keyboard navigation and enable more of the combobox tests
remove VirtualizedList. Not used and if we need this we will use Table
rename a couple of Filter components to avoid confusion with more generic components


vuu-ui/packages/vuu-ui-controls/src/list/useList.ts
```diff
@@ -84,17 +84,12 @@ export const useList = <Item, S extends SelectionStrategy>({
...
-  const scrollContainer = useMemo<RefObject<HTMLElement>>(() => {
-    if (scrollContainerRef) {
-      return scrollContainerRef;
-    } else {
-      return {
-        current:
-          containerRef.current?.querySelector(".vuuList-scrollContainer") ??
-          null,
-      };
+  const setContainerRef = useCallback<RefCallback<HTMLDivElement>>((el) => {
```
heswell, line 87: This is the crux of the List navigation fix. The problem was that some of the navigation functions rely on the containerRef being set. Because of the way list renders, a child of MeasuredItem, the containerRef , defined as a regular RefObject, was only set on second render. If navigation happens immediately, the ref was not yet set.

By using a RefCallback, we ensure that the refs are set immediately the List is rendered.

vuu-ui/packages/vuu-ui-controls/src/list/common-hooks/useKeyboardNavigation.ts
```diff
@@ -108,33 +111,29 @@ const pageDown = (
...
+    let nextRect;
+    do {
+      nextIdx -= 1;
+      nextRect = getElementByDataIndex(
+        containerEl,
+        nextIdx,
+        true
+      ).getBoundingClientRect();
+    } while (nextRect.top > itemTop && nextIdx > 0);
+    return nextIdx;
```
junaidzm13, line 136: Unrelated to this change, but (for the future) the flow would be easier to follow if we can have a separate function to calculate the next index. For example:
```
if (index > 0) {
  if (newScrollTop === scrollTop) {
    return 0;
  } else  {
    containerEl.scrollTo(0, newScrollTop);
    return next-idx-fn(); // should be a better name obv
}
```

junaidzm13 (approved):
Looks good!

Issue #1167, opened 2024-01-29
Handle when ignite data schema is different from VUU table schema


naleeha:
- Currently done using a simple map for sorting - but the name of the columns on ignite and vuu table is hard coded. This needs to referenced from the ignite data schema and VUU table schema to avoid them getting out of sync with whats specified in the map
- Need to support when ignite and vuu column names or the column types differ for filtering.  Currently only works if they are the same

Places where we currently need to know the Ignite field names/types are
- Ignite configuration for ignite server side to specify the schema - can be specified when server start up or at run time.
- When we read the ignite data from the sql field query (see toChildOrder() https://github.com/finos/vuu/blob/main/example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/IgniteOrderStore.scala)
- When we update the table rows with data fetched from ignite (https://github.com/finos/vuu/blob/main/example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/provider/IgniteOrderDataProvider.scala)
- When we need to  translate sort + filter specified on the table to ignite data sort and filter (see ColumnMap in https://github.com/finos/vuu/blob/main/example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/provider/IgniteOrderDataProvider.scala)

naleeha:
Currently we are experimenting with different ways of specifying the data (entity) schema on Ignite server side.

- Using QueryEntity to define the fields + indexes https://github.com/finos/vuu/blob/main/plugin/ignite-plugin/src/test/scala/org/finos/vuu/feature/ignite/IgniteTestStore.scala
- Using Dto object

Looks like the fields defined using QueryEntity has api for getting the field name and values
https://ignite.apache.org/releases/latest/dotnetdoc/api/Apache.Ignite.Core.Cache.Configuration.QueryEntity.html#Apache_Ignite_Core_Cache_Configuration_QueryEntity_ValueFieldName

naleeha:
It would be good if the table schema we define for this can be used for all other usecases https://github.com/finos/vuu/discussions/853

PR #1183, opened 2024-02-06
#1167 handle when external data schema is different from internal 

junaidzm13:
https://github.com/finos/vuu/issues/1167 [handle when external data schema is different from internal](https://github.com/finos/vuu/pull/1183/commits/a2c9c496c2761c7b1a1c97747ed5bb426f687c24) 

- uses SchemaMapper to build the schema mappings b/w
  external and internal.
- potential design/performance tradeoff: inside IgniteOrderStore.findChilOrder
  method toChildOrder can be a performance hit, can keep as raw list as we
  convert back to a list anyways when mapping to row data.
- Sort + Filters to use the mapper to build SQL queries instead
  of internal table def.
- refactors and improves the flow of IgniteOrderDataProvider.


example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/schema/IgniteEntitySchema.scala
```diff
@@ -0,0 +1,95 @@
...
+import scala.reflect.runtime.universe.{MethodSymbol, TypeTag, typeOf}
+
+case class IgniteEntitySchema private (private val columnDef: mutable.LinkedHashMap[ColumnName, IgniteDataType],
+                              queryIndex: List[QueryIndex]) extends ExternalStoreEntitySchema {
+
+  private val _schemaFields: List[SchemaField] = columnDef.zipWithIndex.map(
+    {case ((name, dType), i) => SchemaField(name, dType, i)}
+  ).toList
+
+  override def schemaFields: List[SchemaField] = _schemaFields
```
naleeha, line 21: Does it make sense to make this a val given that we don't expect use case of ExternalStoreEntitySchema be re-evaluated at the time of gettingSchemaFields? 
If we are trying to solve for the fact you are generating the schemaFields at run time, i think we can solve it using companion object or in this case, do it in the IgniteEntitySchemaBuilder
https://stackoverflow.com/questions/24199490/scala-case-class-generated-field-value

example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/module/IgniteOrderDataModule.scala
```diff
@@ -4,31 +4,46 @@ import org.finos.toolbox.lifecycle.LifecycleContainer
...
+    "price" -> ("price", classOf[Double]),
+    "quantity" -> ("quantity", classOf[Int]),
+    "side" -> ("side", classOf[String]),
+    "strategy" -> ("strategy", classOf[String]),
+    "parentId" -> ("parentId", classOf[Int]),
+  ).zipWithIndex
+    .map({ case ((extField, (name, t)), i) => (extField, SimpleColumn(name, i, t)) })
+    .toMap
+
+  val schemaMapper: SchemaMapper = SchemaMapper(IgniteChildOrderEntity.getSchema, tableColumnByExternalField)
```
naleeha, line 48: I think we want to reverse the dependencies. Instead of creating schema mapper and generating/getting table columns, define the table columns and pass that into provider to create a schema mapper with the column mappings
This means the responsibility of knowing table column -> external column stays in the provider.
If you wanted to swap out the provider, you can without changing the table def

naleeha (commented):
Looks good and solves the issue we have with sort and filter
Added some comments for design discussions & clarify intentions.
One future improvements we could do is to create a ignite sql builder that combines all the sql logic in fitler, sort + ignite order store currently - makes it easy to unit test and re-use (the mock to we have in the test to assert call with right sql param for example can be replaced with pure sql building test)

plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/schema/SchemaMapper.scala
```diff
@@ -0,0 +1,54 @@
...
+      val f = externalSchemaField(column.name).get
+      val columnValue = values(f.index) // @todo add type conversion conforming to the passed schema
+      (column.name, columnValue)
+    }).toMap
+  }
+
+  private def getExternalSchemaFieldsByColumnName = {
+    _schemaFields.map(
+      f => if (tableColumnByExternalField.contains(f.name)) (tableColumnByExternalField(f.name).name, f) else null
+    ).filter(v => v != null).toMap
```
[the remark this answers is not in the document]
junaidzm13: Ahh yeah that we can do. Let me update

example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/schema/IgniteEntitySchema.scala
junaidzm13: Yeah I think it does. I've updated it to instead of overriding a def it overrides a val from the extended trait `ExternalStoreEntitySchema`.

That way we don't need to change the `IgniteEntitySchemaBuilder` and pass in two objects to the `IgniteEntitySchema` where one is derived from the other `fieldDefs` vs `schemaFields`.

example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/IgniteOrderStore.scala
```diff
@@ -58,25 +59,17 @@ class IgniteOrderStore(private val parentOrderCache: IgniteCache[Int, ParentOrde
...
-
-    var counter = 0
-    val buffer = mutable.ListBuffer[ChildOrder]()
-    results.forEach(item => {
-      buffer.addOne(toChildOrder(item))
-      counter += 1
-    })
-
-    logger.info(s"Loaded Ignite ChildOrder for $counter rows, from index : $startIndex where $whereClause order by $sqlSortQueries")
+    val results = LazyList.from(childOrderCache.query(query).asScala).map(i => toChildOrder(i.asScala.toList))
```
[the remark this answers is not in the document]
junaidzm13: Not sure if it achieves the same, but would be ideal if we can take advantage of that as well. I'll look into this.

My thinking was mostly around to delay the `List` ->`toChildOrder` -> `toRowData` conversion until actually needed, which in this case would be when we update the table row.

Before we were iterating through the query result and adding it to a ListBuffer, so that wasn't taking advantage of the lazy load from the query cursor anyways.

PR #1185, opened 2024-02-09
#1167 refactor to use FP paradigm for schema validation logic

junaidzm13:


plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/schema/SchemaMapper.scala
```diff
@@ -32,33 +32,22 @@ object SchemaMapper {
...
   private def externalFieldsInMapConformsToExternalSchema(externalSchema: ExternalStoreEntitySchema,
                                                           externalFields: Iterable[String]): ValidationError = {
-    externalFields.foreach(field => {
-      val exists = externalSchema.schemaFields.exists(_.name == field)
-      if (!exists) return Some(s"Field `$field` not found in external schema")
-    })
-
-    None
+    externalFields
+      .find(field => externalSchema.schemaFields.forall(_.name != field))
```
naleeha, line 41: is there a reason to switch to forall rather than using exists?
Exists will exit early on first element it find that is a match rather than traversing the whole collection 
https://stackoverflow.com/questions/59922992/when-true-does-scala-exists-stop-at-the-first-element-that-satisfies-the-predic

plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/schema/SchemaMapper.scala
```diff
@@ -32,33 +32,22 @@ object SchemaMapper {
...
-    )
-
-    if (columns.length > internalFields.size)
-      return Option(s"More internal columns passed than mapped fields")
-
-    None
+    internalFields
+      .find(columnName => columns.forall(_.name != columnName))
+      .map(columnName => s"Column `$columnName` not found in internal columns")
+      .orElse(Option.when(columns.length > internalFields.size)(s"More internal columns passed than mapped fields"))
```
naleeha, line 50: Not sure having no mapping for a column is an error case.
If we look at BasketConstituent example, there are several csv files, one for each basket. It could be valid that one of the basket is missing any values for a BasketConstituent field

plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/schema/SchemaMapper.scala
junaidzm13: `forall` exits early as well, so they're same in that aspect. I just found `forall` easier to follow in this case

example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/schema/IgniteEntitySchema.scala
```diff
@@ -84,9 +84,9 @@ case class IgniteEntitySchemaBuilder(private val fieldDef: mutable.LinkedHashMap
   private type ValidationError = Option[String]
   private def validateSchema = indexAppliedOnAbsentField()
   private def indexAppliedOnAbsentField(): ValidationError = {
-    index.foreach(_.getFieldNames.asScala.foreach(name =>
-      if (!fieldDef.contains(name)) return Some(s"Field `$name` not found in schema.")
-    ))
-    None
+    index.iterator
```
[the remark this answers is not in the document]
junaidzm13: I think we do need it if we want lazy eval and exit as soon as we find an error.

But yeah a different story if we want to support all invalid index :)

plugin/ignite-plugin/src/main/scala/org/finos/vuu/feature/ignite/schema/SchemaMapper.scala
junaidzm13: I was unaware of this edge case. Thanks for pointing this out.

example/apache-ignite/src/test/scala/org/finos/vuu/example/ignite/schema/IgniteEntitySchemaTest.scala
```diff
@@ -33,7 +33,7 @@ class IgniteEntitySchemaTest extends AnyFeatureSpec with Matchers {
           .build()
       )
       exception shouldBe a[RuntimeException]
-      exception.getMessage should include regex "[Ff]ield `name` not found"
+      exception.getMessage should include regex s"[Ff]ield `name` in index `NAME_IDX`.*not found"
```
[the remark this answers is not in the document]
junaidzm13: My initial intention was to return only the first invalid index. Now I've changed it to include all.

Issue #1213, opened 2024-02-20
Hide row creation behind an interface with plugable implementations


chrisjstevo:
## Feature Request

In order to hide the implementation for row creation we'd like to hide the RowWithData object behind a builder interface

PR #1214, opened 2024-02-21
DRAFT #1213 added example of rowbuilder interface

chrisjstevo:


rumakt:
Looking good.

vuu/src/main/scala/org/finos/vuu/core/row/InMemMapRowBuilder.scala
```diff
@@ -0,0 +1,45 @@
...
+    this
+  }
+
+  override def setBoolean(column: Column, v: Boolean): RowBuilder = {
+    mutableMap.put(column.name, v)
+    this
+  }
+  override def setKey(key: String): RowBuilder = {
+    this.key = key
+    this
```
junaidzm13, line 36: Not sure if it applies here but would be worthwhile to think about:

It uses the same instance of the row builder so if somewhere we instantiate a row builder and share the same builder with other functions there is a chance that we might mutate the data in multiple places resulting in unexpected data changes and hard to find bugs.


For example:
```
def main() = {
  ...
  val rowBuilder = InMemMapRowBuilder().setKey("key")
  function1(rowBuilder)
  function2(rowBuilder)
}

def function1(rowBuilder: RowBuilder, ...) = {
  ...
  rowBuilder.setLong(col1, 50).asRow()
  ...
}

def function2(rowBuilder: RowBuilder) = {
  ...
  // key will be "" here, and col1 will have a value of 50
  rowBuilder.setString(col2, "ABC").asRow()
  ...
}
```

vuu/src/main/scala/org/finos/vuu/core/row/InMemMapRowBuilder.scala
```diff
@@ -0,0 +1,45 @@
...
+import org.finos.vuu.core.table.{Column, RowData, RowWithData}
+
+import scala.collection.mutable
+
+class InMemMapRowBuilder extends RowBuilder {
+
+  private val mutableMap = new mutable.HashMap[String, Any]()
+  private var key: String = ""
+  override def setLong(column: Column, v: Long): RowBuilder = {
+    mutableMap.put(column.name, v)
```
junaidzm13, line 11: For the future: might be able to check for type consistency by matching the type i.e. `Long` with `column.dataType`.

junaidzm13 (commented):
Looks good! Left a couple of comments, let me know what you think

vuu/src/main/scala/org/finos/vuu/core/row/InMemMapRowBuilder.scala
chrisjstevo: Yes, that is very true. Each caller should getetter own I think, but I think it's fair to assume different call sites shouldn't pass them across threads.
Will update and add notthreadsafe annotation

vuu/src/main/scala/org/finos/vuu/core/row/InMemMapRowBuilder.scala
chrisjstevo: Yes, that was indeed my goal. So checking datatype of column matches call signature.
I will add that.

PR #1215, opened 2024-02-21
Ignite row count

naleeha:


example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/IgniteOrderStore.scala
```diff
@@ -87,6 +87,16 @@ class IgniteOrderStore(private val parentOrderCache: IgniteCache[Int, ParentOrde
 
   }
 
+  def getCount(sqlFilterQueries: String): Long = {
+    //todo should this be COUNT_BIG?
+    val whereClause = if(sqlFilterQueries == null || sqlFilterQueries.isEmpty) "" else s" where $sqlFilterQueries"
+    val query = new SqlFieldsQuery(s"select COUNT(1) from ChildOrder$whereClause")
+    val cursor = childOrderCache.query(query)
+
+    val countValue = cursor.getAll().get(0).get(0)
```
chrisjstevo, line 96: Would we need to put a guard around this? I guess not as there should always be a count (even if its zero)

example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/IgniteOrderStore.scala
```diff
@@ -87,6 +87,16 @@ class IgniteOrderStore(private val parentOrderCache: IgniteCache[Int, ParentOrde
 
   }
 
+  def getCount(sqlFilterQueries: String): Long = {
+    //todo should this be COUNT_BIG?
+    val whereClause = if(sqlFilterQueries == null || sqlFilterQueries.isEmpty) "" else s" where $sqlFilterQueries"
+    val query = new SqlFieldsQuery(s"select COUNT(1) from ChildOrder$whereClause")
+    val cursor = childOrderCache.query(query)
```
chrisjstevo, line 94: One thing we should think about is whether we'd want to parametrize the sql statement, at the moment I think we'd be vulnerable to a sql injection attack on the where clause. Obv this is not a problem for the demo, but something we might want to clarify from a usage perspective. 

Would be interested in @rumakt 's views on this.

naleeha: On the ignite side? 
On the vuu side the sql filter query is generated from our ui filter query syntax.
Do you mean the interface for this method should not accept a string filter query or worried about attacker sending their own VUU sql syntax via the websocket?

chrisjstevo: I'd be worried about an attacker submitting something like this: 

columnName ='FOO\';DELETE * FROM Orders' in a filter statement and it making its way all the way to Ignite. 

I'm not sure if Ignite supports prepared statements: 

"Select * from Orders Where column = ?"

And we get the driver/client lib to do the software parse, (so that a sql attack can't happen) then we do like this: 

query.setParameter(0, <value from client>);

If that makes sense. Its not a problem for the demo code, but might imply the string concat is ok in prod system.

PR #1217, opened 2024-02-22
added index.html to showcase

chuchee:
added index.html to showcase


heswell (changes requested):
please can you revert the package-lock.json, There should be no need to commit that here

heswell (approved):

Issue #1195, opened 2024-02-14
Need to track metrics for each instance of virtual session table


naleeha:
### Steps to Reproduce:
 1. run ignite vuu 
2. load ignite big order table twice - two instances
3. Load Metics table

### Expected Result:
Two instances of the big order table with long names that start with session:xxx

### Actual Result:
One record for big order with archetype name  (Ignite big order2 ) 

### Additional Context:

Need to be careful to ensure the table menu still show the ignite table archetypes (Ignite big order2 ) and not the virtual session tables
But in Metrics, show the instances of virtual session tables but not the archetype  (Ignite big order2 )

The gettable methods does not distinguish the session table name.

![image](https://github.com/finos/vuu/assets/148964225/b0de741b-4967-4e11-84f3-a6965d615eb5)

PR #1218, opened 2024-02-22
#1195 enable metric tracking for each instance of session table

junaidzm13:
Couple of issues that needed fixing:
  1. TableContainer.getTables() returned array where table name actually
     reflected the tableDef.name. Because of this returned session tables
     and its blueprints had same names and the metrics table only added
     one and it seemed as if we're only considering the archetype.
  2. this one was a bit more involved. First we were using tableDef.name
     to update metrics in InMemDataTable, hence, if we updated metrics for
     a session table it appeared as if we're updating for the archetype.
     Second, if we even had tried calling sessionTable.name in InMemDataTable
     to initialize the metric counters/meters it still would've not worked
     because in session table we used an instance variable `createInstant`
     (used in session-table name) which is not initialized at the time we
     call super class (InMemDataTable) constructor.

Also,
- renames getNonSessionTables to getDefinedTables since its intended
  to return unique table defs for the UI. Also, now it explicitly does that
  by filtering out session tables (i.e. non-unique table defs).
- getTables (for now) is intended to return actual tables and not the blueprints
  for MetricsTable. That's what it does now.

Screenshot 2024-02-23 at 10 44 17 AM

### DIFF
--- a/toolbox/src/main/scala/org/finos/toolbox/lifecycle/LifecycleContainer.scala
+++ b/toolbox/src/main/scala/org/finos/toolbox/lifecycle/LifecycleContainer.scala
@@ -226,7 +226,7 @@ class LifecycleContainer(implicit clock: Clock) extends StrictLogging {
   val thread = new Runner("lifeCycleJoinRunner", () => {Thread.sleep(1000)})
   thread.runInBackground()
 
-  val dependencyGraph = new DirectedAcyclicGraph[LifecycleEnabled]()
+  private val dependencyGraph = new DirectedAcyclicGraph[LifecycleEnabled]()
 
   def autoShutdownHook(): Unit = {
     val container = this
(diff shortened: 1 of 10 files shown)


naleeha:
> Couple of issues that needed fixing:
> 
> 1. TableContainer.getTables() returned array where table name actually
>    reflected the tableDef.name. Because of this returned session tables
>    and its blueprints had same names and the metrics table only added
>    one and it seemed as if we're only considering the archetype.
> 2. this one was a bit more involved. First we were using tableDef.name
>    to update metrics in InMemDataTable, hence, if we updated metrics for
>    a session table it appeared as if we're updating for the archetype.
>    Second, if we even had tried calling sessionTable.name in InMemDataTable
>    to initialize the metric counters/meters it still would've not worked
>    because in session table we used an instance variable `createInstant`
>    (used in session-table name) which is not initialized at the time we
>    call super class (InMemDataTable) constructor.
> 
> Also,
> 
> * renames getNonSessionTables to getDefinedTables since its intended
>   to return unique table defs for the UI. Also, now it explicitly does that
>   by filtering out session tables (i.e. non-unique table defs).
> * getTables (for now) is intended to return actual tables and not the blueprints
>   for MetricsTable. That's what it does now.
> 
> Screenshot 2024-02-23 at 10 44 17 AM

Looks perfect for the virtual tables.
For In memory session table, i think we want different behavior.. 
We don't want to see it in the list of metrics tables.
This sessions for in memory tables are used for holding temporary states related to viewport only - e.g. when we do group by, the tree session table and when user edit, the session table.
The data is written directly into the InMemoryDataTable where else for Ignite BigOrder, only the VirtualSessionTable will have data in it.

@chrisjstevo  what do you think?

image

chrisjstevo:
This change looks good. @naleeha I'm not sure on your point re: InMem vs Virtual, lets catchup on this.

PR #1237, opened 2024-02-27
fix issue with Dropdown cell renderer

heswell:
subtle bug in Dropdown cell renderer. Although updates are always  sent to server, because row key is not checked in the 'should component re-render' checker, sometimes we might have a stale key from another row , when the data is same between two rows and scrolling means the cells have been re-used. In this scenario, update is sent with wrong key value.
Introduce a new checker for editable cells, which includes row key in compare.

Additionally, have removed the state held within component - up[dates applied by user will always arrive back from server - better to wait for updates to arrive that way.

### DIFF
--- a/vuu-ui/packages/vuu-utils/src/column-utils.ts
+++ b/vuu-ui/packages/vuu-utils/src/column-utils.ts
@@ -237,6 +237,8 @@ export function projectUpdates(updates: number[]): number[] {
   return results;
 }
 
+const KEY = 6;
+
 export const metadataKeys = {
   IDX: 0,
   RENDER_IDX: 1,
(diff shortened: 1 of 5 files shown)


vuu-ui/packages/vuu-utils/src/column-utils.ts
```diff
@@ -1210,3 +1214,20 @@ export const dataAndColumnUnchanged = (
...
+ * edits data - ensures we never have a stale key.
+ */
+export const dataColumnAndKeyUnchanged = (
+  p: TableCellRendererProps,
+  p1: TableCellRendererProps
+) =>
+  p.column === p1.column &&
+  p.row[KEY] === p1.row[KEY] &&
+  p.column.valueFormatter(p.row[p.columnMap[p.column.name]]) ===
+    p1.column.valueFormatter(p1.row[p1.columnMap[p1.column.name]]);
```
junaidzm13, line 1233: Could've reused the above `dataAndColumnUnchanged` to avoid duplication. I know it's already merged, just leaving this here for future reference.

heswell: Good point will tidy in one of my next PRs

Issue #1236, opened 2024-02-27
Reverse sort direction


naleeha:
### Steps to Reproduce:
1. Load a table in the ui
2. Sort one of the text field

### Expected Result:
Ascending should return A -Z
Descending should return Z -A

### Actual Result:
Ascending returns Z -A

PR #1240, opened 2024-02-28
#1236 fix sort directions

junaidzm13:
- refactored the related code to make it more readable.
- added targeted tests around the fixes in SortCompares object
- improved existing `GenericSortTest` by separating assertions into separate Scenarios/tests. Would help with debugging / fixing tests in the future.
- also fixed warnings related to `Lower cased long literal marker` in `GenericSortTest`.


junaidzm13:
Screenshot 2024-02-28 at 12 46 10 PM

Issue #1248, opened 2024-03-01
Add data change checks to provider to limit updates sent to the UI


junaidzm13:
### Current behaviour

We are sending `TABLE_UPDATE` web-socket messages on every cycle to the UI even when the row data hasn't changed.

For instance, here the data is static, we haven't scrolled but the UI is being unnecessarily spammed with a lot of WS messages:

![Screenshot 2024-03-01 at 12 04 11 PM](https://github.com/finos/vuu/assets/62522218/27e6fa78-7da7-4eba-b890-03c9531861b3)

### Expected behaviour

Only send updates to the UI when the data has changed.

PR #1249, opened 2024-03-01
#1248 add checks to determine if update needs to be sent

junaidzm13:


junaidzm13:
Screenshot 2024-03-01 at 12 22 38 PM

chrisjstevo:
Hi, 

This looks good, but I have one question. with the big data provider the index of the order can change in a dataset (due to sorting). But they key may not. 

For example if I have 10 rows, and key "A" is at top and key "J" is at bottom, both exist in the table. When a reverse alpha sort them, J is now at index 0, and A at 9. From a key perspective they both still exist, but their indices have changed. 

I think you need to not only check that the key and data has not changed, but for big data you need to make sure index is still pointing to that key also. 

Make sense?

Chris

junaidzm13:
Not sure why, but I can't seem to reply to your comment @chrisjstevo. So replying here:

It totally makes sense, thanks for pointing that out!

I've updated the PR - now I'm checking fetched row at a given index against the row in the table at the same index. Which performs all the required checks. Lmk what you think.

chrisjstevo:
I also can't replay to your comment :-) but looks good. Will merge now.

PR #1255, opened 2024-03-06
1203 add config option to disable ssl

junaidzm13:
Summary of the changes:
  - now we can enable/disable `ssl` in the demo app by setting the property `vuu.ssl` to `true/false`.
  - and to build UI demo so that it uses the insecure websocket connection, need to run `build:app:insecure` or `build:app -- --insecure`. Or if the `vuuConfig` is to be directly used then either specify the `websocketUrl` directly or set the `ssl` property so the fallback/default websocket url reflects the required protocol (`ws` vs `wss`).

### DIFF
--- a/example/main/src/main/scala/org/finos/vuu/SimulMain.scala
+++ b/example/main/src/main/scala/org/finos/vuu/SimulMain.scala
@@ -1,6 +1,6 @@
 package org.finos.vuu
 
-import com.typesafe.config.ConfigFactory
+import com.typesafe.config.{Config, ConfigFactory}
 import com.typesafe.scalalogging.StrictLogging
 import org.finos.toolbox.jmx.{JmxInfra, MetricsProvider, MetricsProviderImpl}
 import org.finos.toolbox.lifecycle.LifecycleContainer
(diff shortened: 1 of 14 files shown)


heswell (approved):
UI code looks good to me

example/main/src/main/scala/org/finos/vuu/SimulMain.scala
```diff
@@ -56,24 +56,25 @@ object SimulMain extends App with StrictLogging {
...
+  private val keyPath = defaultConfig.getString("vuu.keyPath")
+  private val sslEnabled = defaultConfig.getBoolean("vuu.ssl")
 
   val config = VuuServerConfig(
     VuuHttp2ServerOptions()
       //only specify webroot if we want to load the source locally, we'll load it from the jar
       //otherwise
       .withWebRoot(webRoot)
-      .withSsl(certPath, keyPath)
+      .withSsl(sslEnabled, certPath, keyPath)
```
chrisjstevo, line 69: Would it be possible to not call this method at all, or indeed call .withUnencryptedWs()? 
It looks a bit odd calling withSsl, passing in certs, but not using them....

junaidzm13: Decided to not call this when ssl is disabled. Can always add another method to do `withUnencryptedWs or withoutWss` if/when needed. Let me know what you think?

chrisjstevo: Ok, that is fine to get this over the line. Can we add another ticket to make opting out of Ssl explicit pls? 

The reason I say is if you don't have to choose explicitly one or the other, we may well find lots of poeple deploying unencrypted websockets or http connections without really understanding they've done that. 

It would be good to have a project under the examples/ folder that has unencrypted websockets/http and explains in a readme what our intended use of this would be (i.e. only in containerized environments where the ssl termination occurs in a sidecar setup at the pod level.)

As said, am happy to approve for now, but lets create another ticket to track this. 

Chris

junaidzm13: Yeah makes sense. Let me do that

junaidzm13: Done: https://github.com/finos/vuu/issues/1258

PR #1260, opened 2024-03-12
1258 make opting out of ssl explicit for http/websocket options

junaidzm13:
make opting out of ssl explicit for http/websocket options

- for now as a quick fix I've set ssl being enabled by default. So that user
  will always have to either call `withSslDisabled` to disable ssl or
  call `withSsl` to provide cert/key paths.


chrisjstevo:
Looks good, thanks.

PR #1276, opened 2024-03-22
Visual Link on same table and with join table

keikeicheung:


keikeicheung:
@heswell have you tested this - we can come back on this after the conf but want to check whether i can merge it

heswell (approved):
tested locally. Looks good.

PR #1284, opened 2024-04-10
Add playwright test for ButtonVariations

chuchee:


heswell (commented):
You won't have seen the linting and typechecking issues yet as the tasks didn't run until I ran them (because of missing cla agreement)
You can always run these locally as a precaution

`npm run typecheck` for example

heswell (commented):
You're running the showcase using the vite dev server. Thats slow to load and probably why you have to use a wait

Better to build the prod version of showcase once, then run against that

``
npm run shopwcase:build
npm run showcase:preview
``
then the showcase will be running on port 4173

I would add the playwright CI actio to the existing test-ui suite
see the cypress-e2e in that file for an exmaple of an existing test that builds and starts prod version of showcase

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
```
heswell, line 4: just need main

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+  pull_request:
+    branches: [ main, master ]
+jobs:
```
heswell, line 7: I don't think you're actually running showcase, unless I'm missing something
see cypress-e2e foe example

vuu-ui/playwright/tests/button.test.ts
```diff
@@ -0,0 +1,16 @@
+import { test, expect } from '@playwright/test';
+
+test('test', async ({ page }) => {
+    const themes: String[] = ['salt', 'vuu']
```
heswell, line 4: types here need to be string, not String

.github/workflows/test-ui.yml
```diff
@@ -23,6 +23,32 @@ jobs:
...
+        uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+      - name: Install Node & dependencies
+        uses: ./.github/actions/setup-vuu-ui
+      - name: Install dependencies
+        run: npm ci
+      - name: Run Vuu Showcase
+        run: npm run showcase:build && npm run showcase:preview
```
heswell, line 40: I think its quite nice to give the two stepd separate labels, the way the cypress task does
Also, wouldn't it be useful to have the wait here too, to ensire the page is loaded
```
          build: npm run showcase:build
          start: npm run showcase:preview
          wait-on: "http://localhost:4173"
 ```

File: .github/workflows/test-ui.yml [hunk 5]
```diff
@@ -23,6 +23,32 @@ jobs:
...
+      - name: Install Playwright Browsers
+        run: npx playwright install --with deps chromium
```
heswell, line 42: we shouldn't need to use npx, we have playwright in our devDependencies, so it should get installed

npx is only for scenarios where the package you need might not be installed

.github/workflows/test-ui.yml
```diff
@@ -23,6 +23,32 @@ jobs:
...
+      - name: Run Playwright tests
+        run: cd ./vuu-ui/playwright && npx playwright test
```
heswell, line 44: see comment above re npx

vuu-ui/package.json
```diff
@@ -41,6 +41,8 @@
     "test:cypress:local": "cypress open --component --browser chrome",
     "test:cypress:e2e": "cypress run --browser chrome",
     "test:vite": "vitest run",
+    "dev": "npm run dev --port 8080",
```
heswell, line 44: what is this task `dev` intended for ?

Issue #1296, opened 2024-04-17
Support schema mapping when entity type is different from table type


naleeha:
### Description of Problem:

Example when we need this 
- Big decimal, until it is supported https://github.com/finos/vuu/issues/1295
- Table Key expects to be String but entity in data source key are not always string

**Considerations**
- Use cases in Java

Issue #1232, opened 2024-02-26
Review security requirement for ignite query


naleeha:
### Description of Problem:

Prevent any sql injection attack via the UI filter or via the websocket message where we convert the user entered filter to sql query to be sent to ignite

PR #1319, opened 2024-05-01
#1296 use parameterized SQL query builder

junaidzm13:
- this would prevent SQL injection and will also be adding better SQL support for different types i.e. for instance before we had to manually convert values to strings (with or without quotes depending on the type) before using them in SQL query. Now, ignite would do that for us.

Probably covers the following issue as well: #1232


naleeha (approved):
looks really good. very readable and extendable

PR #1388, opened 2024-06-14
503 Typeahead service should be case insensitive

wendymiaoo:
Choose toLowerCase as solution considering Iterator is lazy and filter will be applied to the first 10 elments, which means no big impact to performance.

### DIFF
--- a/vuu/src/test/scala/org/finos/vuu/core/table/InMemColumnValueProviderTest.scala
+++ b/vuu/src/test/scala/org/finos/vuu/core/table/InMemColumnValueProviderTest.scala
@@ -80,6 +80,21 @@ class InMemColumnValueProviderTest extends AnyFeatureSpec with Matchers with Moc
       uniqueValues should contain theSameElementsAs Vector("VOA.L", "VOV.L")
     }
 
+    Scenario("Get all unique value of a given column that starts with specified string case insensitive") {
+      val table = givenTable(pricesDef)
+      val provider = new MockProvider(table)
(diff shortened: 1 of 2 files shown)


naleeha (approved):

vuu/src/test/scala/org/finos/vuu/core/table/InMemColumnValueProviderTest.scala
```diff
@@ -80,6 +80,21 @@ class InMemColumnValueProviderTest extends AnyFeatureSpec with Matchers with Moc
...
+      val table = givenTable(pricesDef)
+      val provider = new MockProvider(table)
+      val columnValueProvider = new InMemColumnValueProvider(table)
+
+      provider.tick("1", Map("id" -> "1", "ric" -> "VOA.L", "bid" -> 220, "ask" -> 223))
+      provider.tick("2", Map("id" -> "2", "ric" -> "BT.L", "bid" -> 500, "ask" -> 550))
+      provider.tick("3", Map("id" -> "3", "ric" -> "VOV.L", "bid" -> 240, "ask" -> 244))
+      provider.tick("4", Map("id" -> "4", "ric" -> null, "bid" -> 240, "ask" -> 244))
+
+      val uniqueValues = columnValueProvider.getUniqueValuesStartingWith("ric", "vo")
```
naleeha, line 93: i would probably add some value with lower case as well for completeness / prevent any regression if someone introduce a bug

PR #1396, opened 2024-06-20
Application settings management

rohinp14:


vuu-ui/packages/vuu-shell/src/application-settings/ApplicationSettingsSchema.json
```diff
@@ -0,0 +1,33 @@
+{
```
heswell, line 1: This belongs in the showcase example as it is test data

vuu-ui/packages/vuu-shell/src/application-settings/ApplicationSettingsPanel.tsx
```diff
@@ -1,58 +1,193 @@
...
+};
+
 const classBase = "vuuApplicationSettingsPanel";
 
-export const ApplicationSettingsPanel = () => {
+export const ApplicationSettingsPanel = ({
+  applicationSettingsSchema,
+  applicationSettings,
+  onApplicationSettingChanged,
+}: ApplicatonSettingsPanelProps) => {
```
heswell, line 173: our props extends A DIVElement (HTMLAttributes<HTMLDivElement>)

that means we potentially have a bunch of additional props here , which we're ignoring. 
We know nthese are all props of a div element so we can spread them stright onto our div.

I like to call the spread prop for this `htmlAttributes`, so it would look like this
```
export const ApplicationSettingsPanel = ({
  applicationSettingsSchema,
  applicationSettings,
  onApplicationSettingChanged,
...htmlAttributes
}: ApplicatonSettingsPanelProps) => {
```
and then you would apply these attributes to the outer div container like this

```
{...htmlAttributes}
```

vuu-ui/packages/vuu-shell/src/application-settings/ApplicationSettingsPanel.tsx
```diff
@@ -1,58 +1,193 @@
...
+
+export interface SettingsSchema {
+  properties: SettingsProperty[];
+}
+
+// Determine the form control type to be displayed
+export function getFormControl(
+  property: SettingsProperty,
+  changeHandler: FormEventHandler,
+  selectHandler: SingleSelectionHandler,
```
heswell, line 40: this is not the right type for the salt dropdown

the selection handler looks like this

```
(evt: SyntheticEvent, newSelected: string[]) => void
```

If you hover over `onSelectinChange` in the line below, intellisense will show you the correct type
```

You can avoid coding the type longhand, if you wish, by directly referencing the original type in the DropdownProps, like this

```
selectHandler: DropdownProps['onSelectionChange']
```


<Dropdown value={currentValue} onSelectionChange={selectHandler}>
```

vuu-ui/packages/vuu-shell/src/application-settings/ApplicationSettingsPanel.tsx
```diff
@@ -1,58 +1,193 @@
...
+  }
+  const dataFieldValue = el.getAttribute("data-field");
+  const saltFormField = document.querySelector(".saltFormField") as HTMLElement;
+  if (saltFormField && saltFormField.dataset.field) {
+    return document.querySelector(`[data-field="${dataFieldValue}"]`);
+  }
+};
+
+// Generates application settings form component
+export const SettingsForm = ({
```
heswell, line 127: there seem to be multiple ways of getting the fieldName. A function like this will work for all cases

```
const getFieldNameFromEventTarget = (evt: SyntheticEvent) => {
  const fieldElement = queryClosest(evt.target, "[data-field]");
  if (fieldElement && fieldElement.dataset.field) {
    return fieldElement.dataset.field;
  } else {
    throw Error("field attribute not defined");
  }
};
```

vuu-ui/packages/vuu-shell/src/application-settings/ApplicationSettingsPanel.tsx
```diff
@@ -1,58 +1,193 @@
...
+  applicationSettingsSchema,
+  applicationSettings,
+  onApplicationSettingChanged,
+}: ApplicatonSettingsPanelProps) => {
+  // Settings Handler for toggle and input buttons
+  const onSettingChanged = useCallback<FormEventHandler>((event) => {
+    const fieldElement = queryClosest(event.target, "[data-field]");
+    const fieldName = getFieldName(fieldElement);
+    onApplicationSettingChanged(fieldName, event.target.value);
+  }, []);
```
heswell, line 137: must not miss the dependencies from the dependencyArray. If you have the ESLint extension installed in VSCode, it should be warning you when these are missing

heswell (approved):

PR #1419, opened 2024-07-04
Input Box Support for User Settings + Panel Tests

rohinp14:
Input box change handler and validation logic for user settings form
Component cypress tests for user settings panel
Add tooltip for user prompts when validation fails


vuu-ui/packages/vuu-shell/src/user-settings/SettingsForm.tsx
```diff
@@ -152,7 +153,7 @@ export type SettingsFormProps = SettingsProps & HTMLAttributes<HTMLDivElement>;
 export const SettingsForm = ({
   settingsSchema: applicationSettingsSchema,
```
heswell, line 154: no need to rename any of these props,. just refer to them in the code as-is, so 
```
export const SettingsForm = ({
settings,
settingsSchema,
onSettingChanged,
...htmlAttributes
})
```

vuu-ui/packages/vuu-shell/src/user-settings/SettingsForm.tsx
```diff
@@ -164,23 +165,33 @@ export const SettingsForm = ({
...
     },
-    [onApplicationSettingChanged]
+    [onSettingChanged]
+  );
+
+  // Change Handler for input boxes
+  const inputHandler = useCallback<FormEventHandler>(
+    (event) => {
+      const fieldName = getFieldNameFromEventTarget(event);
+      const { value } = event.target as HTMLInputElement;
```
heswell, line 191: this will always be a string. WIll need to convert to a numeric, if that is what the property represents

heswell (approved):

vuu-ui/packages/vuu-shell/src/user-settings/SettingsForm.tsx
```diff
@@ -137,22 +138,60 @@ export function getFormControl(
...
-    return <Input value={value} />;
+    const valid = isValidInput(currentValue, property.type);
+    const content = getTooltipContent(property.type, valid);
+    const TooltipProps = {
+      tooltipContent: content,
+    };
+    return (
+      <VuuInput
+        key={property.name}
+        onCommit={inputHandler}
```
heswell, line 149: we're not setting the value here

heswell (changes requested):
For the type issue, in BasketToolbar, change code to

```
     TooltipProps={{ tooltipContent: unitErrorMessage }}
```

heswell (approved):

PR #1431, opened 2024-07-16
Web socket api test for get table meta

naleeha:
- Introduced way to write web socket api test - working progress and will evolve as we write more scenarios
- Introduced configuration for turning of heartbeat for debugging/testing
- fixed error scenarios for getting table meta data


rumakt (approved):

vuu/src/main/scala/org/finos/vuu/net/ViewServerClient.scala
```diff
@@ -50,7 +50,10 @@ class WebSocketViewServerClient(ws: WebSocketClient, serializer: Serializer[Stri
     }
     else {
       Try(serializer.deserialize(msg)) match {
-        case Success(vsMsg) => vsMsg
```
heswell, line 53: is it intentional that we log every message ?

heswell (approved):
looks ok to me, but I'm ill qualified to judge scala code

PR #1440, opened 2024-07-23
Feature/java friendly rpc handling

rumakt:
Added JAVA friendly way of RPC Handler, 

RPC handler can be now created under any name and out of Scala or Java function


vuu/src/main/scala/org/finos/vuu/net/rpc/DefaultRpcHandler.scala
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  override def processRpcCall(msg: ViewServerMessage, rpc: RpcCall)(ctx: RequestContext): Option[ViewServerMessage] = {
+    val method = rpc.method
+    val params = rpc.params
+    val namedPars = rpc.namedParams
+    val module = Option(msg).map(_.module).getOrElse("")
+
+    processRpcMethodHandler(method, params, namedPars, ctx) match {
+      case result: RpcMethodSuccess => Some(VsMsg(ctx.requestId, ctx.session.sessionId, ctx.token, ctx.session.user, RpcResponse(method, result, error = null)))
```
naleeha, line 42: should success case return module as well to match the error case?

vuu/src/main/scala/org/finos/vuu/net/rpc/DefaultRpcHandler.scala
```diff
@@ -0,0 +1,64 @@
...
+      case error: RpcMethodFailure => Some(VsMsg(ctx.requestId, ctx.session.sessionId, ctx.token, ctx.session.user, RpcResponse(rpc.method, null, Error(error.error, error.code)), module))
+    }
+  }
+
+  private def processRpcMethodHandler(methodName: String, params: Array[Any], namedParams: Map[String, Any], ctx: RequestContext) = {
+    if (methodHandlers.containsKey(methodName)) {
+      methodHandlers.get(methodName) match {
+        case null => new RpcMethodFailure(s"Could not find rpcMethodHandler $methodName")
```
naleeha, line 50: this will never execute as if there is no matching methodName, it would have gone to line 61 based on the containsKey check result.

vuu/src/main/scala/org/finos/vuu/net/rpc/RpcMethodHandler.scala
```diff
@@ -0,0 +1,27 @@
...
+  override def call(rpcParams: RpcParams): RpcMethodCallResult = handler(rpcParams)
+}
+
+class RpcFunctionMethodHandler(handler: RpcParams => RpcMethodCallResult) extends RpcMethodHandler {
+  override def call(rpcParams: RpcParams): RpcMethodCallResult = handler(rpcParams)
+}
+
+class RpcParams(val params: Array[Any], val namedParams: Map[String, Any], ctx: RequestContext)
+
+trait RpcMethodCallResult {}
```
naleeha, line 19: Can be a separate discussion but looks like this usecase is a good candidate for reviewing the Result Pattern discussion

vuu/src/test/scala/org/finos/vuu/net/rpc/DefaultRpcHandlerTest.scala
```diff
@@ -0,0 +1,85 @@
...
+
+class DefaultRpcHandlerTest extends AnyFlatSpec with Matchers with BeforeAndAfterEach {
+  var handler: DefaultRpcHandler = _
+  val ctx = RequestContext("requestId", ClientSessionId("sessionId", "user"), null, "token")
+
+  override def beforeEach(): Unit = {
+    handler = new DefaultRpcHandler
+  }
+
+  "The registerRpcMethodHandler method" should "register a handler for a given rpc method" in {
```
naleeha, line 17: All existing tests use AnyFeatureSpec. We don't have actual tests that use AnyFlatSpec.
Official guide mentions using AnyFlatSpec as default and  AnyFeatureSpec for more acceptance style
https://www.scalatest.org/user_guide/selecting_a_style

Let's have a quick decision as a team which style(s) we want to use + when so we can try and keep things more consistent

rumakt: This is not acceptance test TBH, valid point though on following already established pattern - rewritten tests as requested

vuu-java/src/test/java/org/finos/vuu/net/rpc/RpcMethodHandlerTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+        final DefaultRpcHandler defaultRpcHandler = new DefaultRpcHandler();
+        defaultRpcHandler.registerRpcMethodHandler("helloWorld", handler);
+
+        RpcCall call = new RpcCall("service", "helloWorld", new Object[]{}, ScalaCollectionConverter.toScala(Collections.emptyMap()));
+        Option<ViewServerMessage> response = defaultRpcHandler.processRpcCall(createRandomViewServerMessage(new LoginRequest("token", "user")), call, ViewPortTestUtils.requestContext());
+
+        Assertions.assertThat(response.get().body())
+                .isExactlyInstanceOf(RpcResponse.class)
+                .isEqualTo(new RpcResponse("helloWorld", new RpcMethodSuccess("It Works"), null));
```
naleeha, line 47: I think we should avoid exposing RpcMethodSuccess object on the RcpResponse.

naleeha: In terms of current use case
- login request currently doesnt go through rpc handler, its special rcp message that get handled at central place. - not an issue in terms of unit test, just fyi
- The only current use case of RcpCall -> RcpResponse is the typeahead. Which returns list of values that match the search criteria as string. I guess idea is that each rcp call can return a different result type? So this change essentially will change the response returned from typeahead.
- We plan to change the type ahead to use ViewPortRpcCall -> ViewPortRpcResponse which returns a ViewPortAction. We dont currently have an action that lets you retunr values like we need for type ahead. Not relevant to this change but just fyi as well

rumakt: OK, not exposing MethodSuccess to the response, it will be consistent with the existing API,

When it comes to the further comments, login is just an example message to not create a new one.

naleeha (approved):

heswell (approved):

PR #1602, opened 2025-07-08
Add ability to disable WebRoot entirely

stonesmi:
Closes https://github.com/finos/vuu/issues/1605

### DIFF
--- a/example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/IgniteVuuMain.scala
+++ b/example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/IgniteVuuMain.scala
@@ -17,7 +17,7 @@ import org.finos.vuu.core.module.vui.VuiStateModule
 import org.finos.vuu.example.ignite.loader.IgniteOrderGenerator
 import org.finos.vuu.example.ignite.module.IgniteOrderDataModule
 import org.finos.vuu.net.auth.AlwaysHappyAuthenticator
-import org.finos.vuu.net.http.VuuHttp2ServerOptions
+import org.finos.vuu.net.http.{AbsolutePathWebRoot, VuuHttp2ServerOptions}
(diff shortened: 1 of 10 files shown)


example/apache-ignite/src/main/scala/org/finos/vuu/example/ignite/IgniteVuuMain.scala
```diff
@@ -63,12 +63,8 @@ object IgniteVuuMain extends App with StrictLogging {
 
   val config = VuuServerConfig(
     VuuHttp2ServerOptions()
-      //only specify webroot if we want to load the source locally, we'll load it from the jar
-      //otherwise
-      .withWebRoot(webRoot)
+      .withWebRoot(AbsolutePathWebRoot(webRoot, directoryListings = true))
```
heswell, line 66: I know its always been this way, but do we actually want directoryListings true

I assume this is what it sounds like and allows user to enter  
https://localhost:8443 
and see directory listing for public .

Alternative is we expect them to use one of

- https://localhost:8443/index.html
- https://localhost:8443/login.html
- https://localhost:8443/demo.html

vuu/src/main/scala/org/finos/vuu/net/http/VuuHttp2ServerOptions.scala
```diff
@@ -9,32 +9,34 @@ trait VuuSecurityOptions {
...
-    VuuHttp2ServerOptionsImpl(port = 8080, allowDirectoryListings = false)
+    VuuHttp2ServerOptionsImpl(port = 8080)
   }
 }
 
 private case class VuuHttp2ServerOptionsImpl(sslEnabled: Boolean = true,
                                              certPath: String = "", keyPath: String = "",
-                                             webRoot: String = "", port: Int,
-                                             allowDirectoryListings: Boolean,
+                                             webRoot: VuuHttp2ServerWebRootType = WebRootDisabled(), port: Int,
```
stonesmi, line 39: @heswell, default behaviour is no web root at all

PR #1621, opened 2025-07-21
Fix viewport typeahead

tanjjj:
Fix viewport typeahead so it ignores the rows that are filtered out in viewport

### DIFF
--- a/vuu/pom.xml
+++ b/vuu/pom.xml
@@ -209,12 +209,6 @@
                 <!--scope>provided</scope-->
         </dependency>
 
-        <dependency>
-            <groupId>org.awaitility</groupId>
-            <artifactId>awaitility-scala</artifactId>
-            <version>${awaitility.version}</version>
-        </dependency>
-
     </dependencies>
 
     <profiles>
--- a/vuu/pom.xml
+++ b/vuu/pom.xml
@@ -209,6 +209,12 @@
                 <!--scope>provided</scope-->
         </dependency>
 
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility-scala</artifactId>
+            <version>${awaitility.version}</version>
+        </dependency>
+
     </dependencies>
 
     <profiles>
--- a/vuu/src/main/scala/org/finos/vuu/net/rpc/Rpc.scala
+++ b/vuu/src/main/scala/org/finos/vuu/net/rpc/Rpc.scala
@@ -1,5 +1,6 @@
 package org.finos.vuu.net.rpc
 
+import org.finos.vuu.feature.ViewPortKeys
 import org.finos.vuu.net.RequestContext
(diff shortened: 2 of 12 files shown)


heswell:
I'll raise a separate request to change the RPC_CALL to a VIEW_PORT_RPC_CALL for typeahead queries from UI

vuu/src/main/scala/org/finos/vuu/net/rpc/Rpc.scala
```diff
@@ -12,6 +13,7 @@ class RpcParams(
                  val params: Array[Any],
                  val namedParams: Map[String, Any],
                  val viewPortColumns: Option[ViewPortColumns],
+                 val vpKeys: ViewPortKeys,
```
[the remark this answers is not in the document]
tanjjj: agreed. Updated now

vuu/pom.xml
```diff
@@ -209,6 +209,12 @@
                 <!--scope>provided</scope-->
         </dependency>
 
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility-scala</artifactId>
+            <version>${awaitility.version}</version>
```
stonesmi, line 215: scope test needed here

vuu/src/test/scala/org/finos/vuu/wsapi/TypeAheadWSApiTest.scala
```diff
@@ -238,10 +242,11 @@ class TypeAheadWSApiTest extends WebSocketApiTestBase {
   }
 
   private def createViewPort = {
-    val createViewPortRequest = CreateViewPortRequest(ViewPortTable(tableName, moduleName), ViewPortRange(1, 100), columns = Array("Id", "Name", "Account"))
+    val createViewPortRequest = CreateViewPortRequest(ViewPortTable(tableName, moduleName), ViewPortRange(1, 100), columns = Array("Id", "Name", "Account"), SortSpec(List(SortDef("Account", 'A'))), Array.empty[String], FilterSpec("Account > 20000"))
     vuuClient.send(sessionId, tokenId, createViewPortRequest)
     val viewPortCreateResponse = vuuClient.awaitForMsgWithBody[CreateViewPortSuccess]
     val viewPortId = viewPortCreateResponse.get.viewPortId
+    await pollDelay(100, MILLISECONDS) until { () => true } // Wait for viewPortRunner to populate view port keys
```
stonesmi, line 249: is there not a way we can check whether the keys have been populated here?

tanjjj: we could wait for another message of table size from vuu to the vuuClient here and confirm the keys are populated. Let me give it a try

stonesmi (approved):

PR #1631, opened 2025-07-24
TRR: Update third party dependencies

### LINKED ISSUE #1606: Review and remediate Vuu runtime dependencies
Vuu currently defines logback as a runtime dependency, however the choice of logging implementation should be up to the consuming application.

Update Vuu to only declare the use of slf4j-api publicly.

Tests and examples can continue to use logback.

We should use this an opportunity to review and remediate any other dependencies with issues or that we don't intend to keep using.

stonesmi:
Will close #1606 

Plan of attack:

* Remove all dependencies from parent-pom
* Align all dependency versions with spring boot 3.5.3 BOM where possible
* Bump bytecode version to Java 17 (required by some updated libs)
* Upgrade anything else with a CVE 
* Replace joda-time with java.time (this isn't 2015!)
* Remove lucene as it isn't used
* Remove logback from the compile scope for libraries.
* Replace guava cache (EOL) with https://github.com/ben-manes/caffeine
* Move benchmark code into benchmarks project to remove JMH from Vuu library(and ported it to Java because the maven JavaDoc plugin is being awkward about mixing Scala and Java in the same module)


stonesmi:
Old dependencies:


New dependencies:

image

tanjjj (approved):

PR #1648, opened 2025-08-01
FilterClause component changes

anandBalachandar:
The filterclause component opens up the suggestions dropdown by default. 
In the scenario where we place multiple FilterClauses (without a filtereditor) on the page, they render with open dropdowns. This dropdown behaviour can now be controlled via a prop.

Also, for multiselect setup, if the values are cleared, the model will trigger a commit on close of the dropdown. (useFilterClause.ts -> handleOpenChange). This will enable the consumer to attach a handler based on the value of the filter.

image

### DIFF
--- a/vuu-ui/packages/vuu-filters/src/__tests__/__component__/FilterClause.cy.tsx
+++ b/vuu-ui/packages/vuu-filters/src/__tests__/__component__/FilterClause.cy.tsx
@@ -3,7 +3,9 @@ import {
   NewFilterClause,
   PartialFilterClauseColumnAndOperator,
   PartialFilterClauseColumnAndOperatorWithDataSource,
-  PartialFilterClauseColumnAndOperatorWithoutDefaultDropdown,
+  FilterColumnAndOperatorWithDropdownOpenOnFocusDisabled,
+  FilterColumnWithDropdownOpenOnFocusDisabled,
+  NewFilterClauseWithDropdownOpenOnFocusDisabled,
 } from "../../../../../showcase/src/examples/Filters/FilterClause/FilterClause.examples";
 
 describe("FilterClause", () => {
@@ -35,7 +37,7 @@ describe("FilterClause", () => {
     });
   });
 
-  describe("WHEN partial filter clause with Column and Operator is rendered and dataSOurce is available", () => {
+  describe("WHEN partial filter clause with Column and Operator is rendered and dataSource is available", () => {
     it("THEN component is rendered with controls for column, operator and value, value suggestions are offered", () => {
       cy.mount(
         <LocalDataSourceProvider>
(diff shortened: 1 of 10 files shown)


vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
```diff
@@ -102,16 +105,17 @@ export const ExpandoCombobox = forwardRef(function ExpandoCombobox<
       ...inputPropsProp,
       onFocus: (evt) => {
         inputPropsProp?.onFocus?.(evt);
+        if (!defaultDropdown && !touched) {
```
heswell, line 108: I don't think the touched concept is working fully as intended. I understand the intention = disable dropdown on initial focus, but once user starts interacting with control, apply auto dropdown on entry as usual.
The reason it isn't working is that as soon as user backspaces and removes operator, the value expando is removed. When operator is selected, a brand new value expando is created, with brand new state, so back to untouched. 
This can easily be fixed, lift the `touched` state (using a ref, not state) up to the FilterClause level (injecting an onFocus to the inputProps) and use this to force showDefaultDropDown to true once touched

vuu-ui/packages/vuu-filters/src/filter-clause/ColumnPicker.tsx
```diff
@@ -10,6 +10,7 @@ export type ColumnPickerProps = Pick<
 > & {
   columns: ColumnDescriptor[];
   onSelect: (evt: SyntheticEvent, columnName: string) => void;
+  defaultDropdown?: ExpandoComboboxProps["defaultDropdown"];
```
heswell, line 13: We show dropdown on focus. Salt used to have the same concept, the older DropDownBase used to have a `openOnFocus` prop, defaulted to true. The current version no longer supports it, but I think its a good descriptive name for the purpose of this prop. `defaultDropdown` doesn't quite describe the purpose of the prop

vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
```diff
@@ -21,6 +21,7 @@ const classBase = "vuuExpandoCombobox";
 export interface ExpandoComboboxProps<Item = string>
```
heswell, line 21: There are no cypress tests for ExpandoComboBox because its such a thin layer round combobox (not that its a good reason there should really be some)
There are some showcase examples, though. IT would be good to have one example that shows the new prop in use

vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
```diff
@@ -63,6 +65,7 @@ export const ExpandoCombobox = forwardRef(function ExpandoCombobox<
   const [value, setValue] = useState(
     valueProp === undefined ? "" : valueProp.toString(),
   );
+  const [touched, setTouched] = useState<boolean>(false);
```
heswell, line 68: nit: don't really need the explicit typing here, it will infer the correct type from the value

vuu-ui/packages/vuu-filters/src/__tests__/__component__/FilterClause.cy.tsx
```diff
@@ -49,4 +50,20 @@ describe("FilterClause", () => {
       cy.findAllByRole("option", { name: "GBP" }).should("be.visible");
     });
   });
+
```
heswell, line 53: would be good to add a test to demonstrate the `touched` so focus into the control, assert that no dropdown renders backspace to remove operator, select an operator, then show that dropdown now appears autotmatically

vuu-ui/packages/vuu-filters/src/filter-clause/FilterClause.tsx
```diff
@@ -15,20 +15,28 @@ import { FilterClauseValueEditor } from "./value-editors/FilterClauseValueEditor
...
 import { OperatorPicker } from "./OperatorPicker";
+import { ExpandoComboboxProps } from "./ExpandoCombobox";
 
 export type FilterClauseCancelType = "Backspace" | "Escape";
 export type FilterClauseCancelHandler = (
   filterClause: FilterClauseModel,
   reason: FilterClauseCancelType,
 ) => void;
 
+export type ShowDefaultDropdown = {
```
heswell, line 26: see comment above I'd suggest
```
showOnFocus: {
columnPicker?: boolean;
operatorPicker?: boolean;
valueEditor?: boolean
}
```

vuu-ui/packages/vuu-filters/src/filter-clause/ExpandoCombobox.tsx
```diff
@@ -21,6 +21,7 @@ const classBase = "vuuExpandoCombobox";
 export interface ExpandoComboboxProps<Item = string>
   extends ComboBoxProps<Item> {
   itemToString?: (item: Item) => string;
+  defaultDropdown?: boolean | undefined;
```
heswell, line 24: nit: its already marked as optional, doesn't need the undefined here

vuu-ui/packages/vuu-filters/src/filter-clause/FilterClause.tsx
```diff
@@ -82,6 +91,7 @@ export const FilterClause = ({
         onSelect={onSelectColumn}
         ref={columnRef}
         value={filterClauseModel.column ?? ""}
+        defaultDropdown={showDefaultDropdown?.forColumnPicker ?? true}
```
heswell, line 94: nit: no need to default true here, it already defaults to true in the component

heswell (changes requested):
almost ! Functionality is mostly good, just one issue - the `touched` detection is isn't always working as intended. This can be fixed easily enough, see comments

vuu-ui/packages/vuu-filters/src/__tests__/__component__/FilterClause.cy.tsx
```diff
@@ -49,4 +52,52 @@ describe("FilterClause", () => {
...
+        cy.mount(
+          <LocalDataSourceProvider>
+            <NewFilterClauseWithDropdownOpenOnFocusDisabled />
+          </LocalDataSourceProvider>,
+        );
+        const container = cy.findByTestId("filterclause");
+        container.find(".vuuFilterClauseField").should("have.length", 1);
+        cy.findByTestId("filterclause")
+          .find(".vuuFilterClauseColumn input")
+          .should("not.be.focused");
```
anandBalachandar, line 68: Do we want keep this in focus?

heswell (approved):

PR #1700, opened 2025-08-31
task/extend-column-filters

anandBalachandar:
Initial changes 
- handle "between" op
- add controlled mode support


vuu-ui/packages/vuu-filters/src/column-filter/ColumnFilter.tsx
```diff
@@ -80,37 +102,48 @@ export const ColumnFilter = ({
...
+              {op}
             </Button>
           </MenuTrigger>
           <MenuPanel>
-            <MenuItem>=</MenuItem>
-            <MenuItem>!=</MenuItem>
-            <MenuItem>starts</MenuItem>
-            <MenuItem>ends</MenuItem>
-            <MenuItem>contains</MenuItem>
+            <MenuItem onClick={() => onOperatorChange("=")}>=</MenuItem>
```
heswell, line 109: The menuitems for operator need to be built dynamically as they will be different, depending on data type of column. See the existing `OperatorPicker` component, which used getOperators. to do exactly this.

heswell:
there is a commit history here including changes that are not related to these files. Can you keep our main branch up to date with finos and regularly  rebase your feature branch onto main

heswell (approved):

PR #1703, opened 2025-09-02
#1702 add invisible flag to table definition

### LINKED ISSUE #1702: Support invisible tables
Problem: not all tables should be visible to users. We may have some tables used internally and containing data that shouldn't be seen by users.

Todo:
add a flag to table definition to mark a table as invisible.
When handling create view port request, if the table is invisible, return the same error message as table doexn't exist.
Allow backend to set a table as invisible when creating table definition.
Double check if we need to reject rpc request/http request

tanjjj:
Add invisible flag to table definition
Reject creating VP if table is invisible
Start using dedicated reject message

Note only supporting invisible flag for standard non-join table for now. I had a battle with JoinTableDef but because joins: JoinTo* is varargs I coulnd't find an easy way to add invisible flag to join table def without changing all lines calling it.

closes #1702


heswell (approved):
looks ok to me

PR #1704, opened 2025-09-03
Column filter changes

anandBalachandar:
- Switched the base interface extension around - Defined the base interface in the hook and extended it in the Component
- Added NumericRange example
- Amended filter value type used in ColumnFilter component


vuu-ui/packages/vuu-data-react/src/data-editing/get-data-item-edit-control.tsx
```diff
@@ -53,6 +53,7 @@ export const getDataItemEditControl = ({
   if (dataDescriptor.editable === false) {
     return (
       <VuuInput
+        data-edit-control="data-edit-control"
```
heswell, line 56: you could just use `data-edit-control` it will default to true, as we're just using it as a flag, that would work.

vuu-ui/packages/vuu-filters/src/column-filter/useColumnFilter.ts
```diff
@@ -128,35 +126,32 @@ export const useColumnFilter = ({
...
 
   const handleInputChange = useCallback<ChangeEventHandler<HTMLInputElement>>(
     (e) => {
       if (Array.isArray(filterValue.current)) {
-        const input = queryClosest(e.target, ".saltPillInput");
-        const updated: [FilterValue, FilterValue] = [
-          (!input?.className?.includes("rangeHigh")
+        const editControl = queryClosest(e.target, "[data-edit-control]", true);
+        const updated: FilterValue = [
+          !editControl?.className?.includes("rangeHigh")
```
heswell, line 143: you don't need the ?. nullable operator after editControl

because you passed true to queryClosesest, it will throw if not found. The types guarantee that response is indeed an element

heswell (approved):

vuu-ui/packages/vuu-filters/src/column-filter/useColumnFilter.ts
anandBalachandar: Updated.

Issue #1753, opened 2025-09-16
Slow filtering performance on join tables


stonesmi:
Filtering and sorting join tables is noticeably slower than on simple data tables.

PR #1754, opened 2025-09-17
Join table performance improvements

stonesmi:
Related to #1753 

* Use a ListBuffer when constructing ViewPortColumns
* Added utility methods backed by lazy vals into ViewPortColumns to avoid repeated calculations and transformations in JoinTable.


stonesmi:
Moving back to draft as I see further improvements

PR #1843, opened 2025-10-24
feat: redesign the freeze control with toggle buttons and a new banne…

BenPreston:
Adds Freeze functionality with two new reusable components. The first is the controls to freeze the table (already exsited but changed) the second shows the frozen banner. Main changes include:

1. **array-data-source**: changes to match the new schema change of vuuUpdatedTimestamp to view the files
2. **FreezeControl**: changed salt switches for toggles, created a new state for flashing. This is so the update badge gives a consistent flash rather than flashing on update as this could be very visually annoying. It also times out three seconds after the last update.
3. **FreezeControl.css**: set animation here for the flashing state using keyframes
4. **FrozenBanner**: created a self contained banner that subscribes to the data source freeze. This is to be used alongside the freeze to appear above the table
5. **FrozenBanner.css**: really minimal style sheet
6. **FreezeTable.example**: demo'd the example by slicing the table to 5 values 
7. **UseFreezeControl**: removed date work as goes in the banner

You can see a GIF demo below:

### DIFF
--- a/vuu-ui/packages/vuu-table-extras/src/__tests__/__component__/freeze-control/FreezeControl.playwright.test.tsx
+++ b/vuu-ui/packages/vuu-table-extras/src/__tests__/__component__/freeze-control/FreezeControl.playwright.test.tsx
@@ -1,20 +1,60 @@
-import { test, expect } from "@playwright/test";
+import { test, expect } from "@playwright/experimental-ct-react";
+import { FreezeControl } from "../../../freeze-control/FreezeControl";
+import React from "react";
+import type { DataSource } from "@vuu-ui/vuu-data-types";
+import { LocalDataSourceProvider } from "@vuu-ui/vuu-data-test";
 
-test("has title", async ({ page }) => {
-  await page.goto("https://playwright.dev/");
-
-  // Expect a title "to contain" a substring.
-  await expect(page).toHaveTitle(/Playwright/);
+const createMockDataSource = (isFrozen = false): Partial<DataSource> => ({
+  isFrozen,
+  table: { module: "TEST", table: "test" },
+  freeze: () => void 0,
+  unfreeze: () => void 0,
+  subscribe: async () => void 0,
+  unsubscribe: () => void 0,
 });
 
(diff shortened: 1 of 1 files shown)


vuu-ui/packages/vuu-table-extras/src/freeze-control/FrozenBanner.tsx
```diff
@@ -0,0 +1,64 @@
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { HTMLAttributes, useEffect, useState } from "react";
+import cx from "clsx";
+import frozenBannerCss from "./FrozenBanner.css";
```
heswell: could you move the css import after the ts imports (with a space between)

vuu-ui/packages/vuu-table-extras/src/freeze-control/useFreezeControl.ts
```diff
@@ -3,46 +3,31 @@ import {
...
-    : "",
   newRecordCount,
 });
 
 export const useFreezeControl = ({ dataSource }: FreezeProps) => {
   const { VuuDataSource } = useData();
   const table = useMemo<VuuTable>(() => {
     if (dataSource.table === undefined) {
-      throw Error(`[useFreezeControls] dataSource must have VuuTable`);
+      throw new Error(`[useFreezeControls] dataSource must have VuuTable`);
```
heswell: there is no difference between new Error() and just Error()

heswell (changes requested):
looks good, one fix and a few minor changes.

A few playwright tests would complete the PR

vuu-ui/packages/vuu-table-extras/src/freeze-control/FrozenBanner.tsx
BenPreston: Sure ... could I do a small follow on task to add https://eslint.org/docs/latest/use/configure/plugins to set this as a custom rule?

PR #1868, opened 2025-11-04
feat: limit freeze control badge to 99 with a + sign

BenPreston:
The Freeze Control Badge wasn't limited so it could grow to any number and this was not helpful and also put the CSS at risk.

This PR limits it to 99 with a + sign to indicate it's over this.


heswell (commented):
do the playwright tests work for you ? They don't work when I run them locally and they don't work in CI. In particular, this funny business ...
```
import * as useFreezeControlModule from "../../../freeze-control/useFreezeControl";
```

causes this error

```
SyntaxError: Named export 'messageHasSize' not found. The requested module '@vuu-ui/vuu-utils' is a CommonJS module, which may not support all module.exports as named exports.
```

heswell (approved):

PR #1898, opened 2025-11-19
Fully implement EpochTimestamp type

### LINKED ISSUE #1643: Implement EpochTimestamp datatype
As per the discussion on https://github.com/finos/vuu/discussions/1089, we seem to generally agree on adding a new type: EpochTimestamp.

This is to represent epoch timestamps only and is not expected to be used for LocalDates or LocalTimes.

stonesmi:
Closes #1643

Migrate the default columns to the new EpochTimestamp type.

As discussed with Chris, changed the precision of EpochTime to milli, to avoid breaking the UI.


stonesmi:
@heswell this is ready for integration on the UI side.

I have made the two default columns of type EpochTimestamp for you to test filtering on.

PR #1927, opened 2025-12-08
1925 Fix incorrect default column values in join tables

### LINKED ISSUE #1925: Fix vuu created and updated timestamp in join table
### Steps to Reproduce:
Use JoinsOfJoinsTableTest as example.
Create a join table, add a few rows. Delete a row. And when debugging we can see in JoinTable.processDelete , the index of keyToIndexMap is updated, but not indexToCreatedTime or indexToLastUpdatedTime. This means the timestamps are not showing correctly for each row.

Related to https://github.com/finos/vuu/issues/1909

tanjjj:
closes #1925 

Removed InnerJoin and use the timestamp from base table instead of the new timestamps for the join table.
Default columns for a join table is now passed in as joinColumns.

### DIFF
--- a/vuu/src/main/scala/org/finos/vuu/api/TableDef.scala
+++ b/vuu/src/main/scala/org/finos/vuu/api/TableDef.scala
@@ -270,11 +270,8 @@ case class JoinTableDef(
                          override val defaultSort: SortSpec,
                          joins: JoinTo*)
   extends TableDef(name, baseTable.keyField, joinColumns, joinFields, indices = Indices(), autosubscribe = false,
-    visibility = visibility, permissionFunction = permissionFunction, defaultSort = defaultSort)
+    visibility = visibility, includeDefaultColumns = false, permissionFunction = permissionFunction, defaultSort = defaultSort)
     with VuuInMemPluginLocator {
-
-  private val defaultColumnsAsJoinColumns: Array[Column] = if (includeDefaultColumns) Columns.fromColumns(this.baseTable, DefaultColumn.getDefaultColumns(customColumns)) else Array.empty
-  private val columns: Array[Column] = if (includeDefaultColumns) customColumns ++ defaultColumnsAsJoinColumns else customColumns
 
   lazy val joinTableColumns = getJoinDefinitionColumnsInternal()
   lazy val rightTables = joins.map(join => join.table.name).toArray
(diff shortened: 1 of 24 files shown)


vuu/src/main/scala/org/finos/vuu/core/table/JoinTable.scala
```diff
@@ -452,7 +445,10 @@ class JoinTable(val tableDef: JoinTableDef, val sourceTables: Map[String, DataTa
     if (keysByTable == null || !keyExistsInLeftMostSourceTable(key))
       EmptyRowData
     else {
-      val columnsByTable = viewPortColumns.getJoinColumnsByTable
+      var columnsByTable: Map[String, List[JoinColumn]] = viewPortColumns.getJoinColumnsByTable
+      if (includeDefaultColumns) {
```
stonesmi: can we move this inside vp columns with a lazy val for performance reasons?

vuu/src/main/scala/org/finos/vuu/api/TableDef.scala
```diff
@@ -223,7 +223,8 @@ class TableDef(val name: String,
                val permissionFunction: (ViewPort, TableContainer) => PermissionFilter = (_, _) => AllowAllPermissionFilter,
                val defaultSort: SortSpec = SortSpec(List.empty)) extends VuuInMemPluginLocator {
 
-  private val columns: Array[Column] = if (includeDefaultColumns) DefaultColumn.addDefaultColumns(customColumns) else customColumns
+  val defaultColumns: Array[Column] = if (includeDefaultColumns) DefaultColumn.getDefaultColumns(customColumns) else Array.empty
```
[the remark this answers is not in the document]
tanjjj: this is changed to private and we no longer need it outside this class

vuu/src/main/scala/org/finos/vuu/api/TableDef.scala
```diff
@@ -281,13 +279,18 @@ case class JoinTableDef(
     visibility = visibility, permissionFunction = permissionFunction, defaultSort = defaultSort)
     with VuuInMemPluginLocator {
 
+  private val defaultColumnsAsJoinColumns: Array[Column] = if (includeDefaultColumns) Columns.fromColumns(this.baseTable, DefaultColumn.getDefaultColumns(customColumns)) else Array.empty
```
stonesmi: We could allow the developers to pass in the default columns and only call DefaultColumn.getDefaultColumns(customColumns) to add them if they aren't in the joinColumns already?

stonesmi (approved):

PR #2041, opened 2026-02-04
Allowing filter name within the Save/Rename Filter prompt to be limited

anandBalachandar:


vuu-ui/packages/vuu-filters/src/filter-provider/FilterContext.ts
```diff
@@ -53,6 +58,7 @@ export interface FilterContextProps {
     key: string,
     filter: string | FilterContainerFilter,
   ) => void;
+  promptInputProps?: PromptInputProps;
```
heswell: There are risks  in offering all the inputProps as overrideable props to client.
Do we really want them to be able to override anything in the input ?
Because its an object, and because it ends up in the dependency array of our useMemo in FilterProvider, users may define it inline in JSX (as the showcase example does) This means it passes an unstable reference which then changes the behaviour of that useMemo. Would be much safer to just accept a `filterNameMaxLength` prop then inject it into the input props at the appropriate point.

Issue #2019, opened 2026-01-29
Memory leak in VuuJoinTableProvider


stonesmi:
### Steps to Reproduce:
Create a join table.
Insert a row in the right hand side table.
Add and remove rows joined to that right side row, which different left primary keys.

### Expected Result:
References to removed rows should not be present in the RightToLeftKeys data structure.

### Actual Result:
References to removed rows are present in the RightToLeftKeys data structure.

### Environment:
Vuu 0.18.0

PR #2055, opened 2026-02-10
#2019 Fixed memory leak when updating the foreign key column in the left table of a join

tanjjj:
This PR fixes the scenario when a row is updated in the left table and that breaks the join, then we want to remove the old mapping of left key - right key from the maps.

### DIFF
--- a/vuu/src/test/scala/org/finos/vuu/core/table/JoinDataTableDataTest.scala
+++ b/vuu/src/test/scala/org/finos/vuu/core/table/JoinDataTableDataTest.scala
@@ -24,6 +24,7 @@ class JoinDataTableDataTest extends AnyFeatureSpec with Matchers {
   val orderId = "123456789"
   val orderId2 = "123456788"
   val orderId3 = "123456787"
+  val orderId4 = "123456786"
   val instrumentRic = "VOD.L"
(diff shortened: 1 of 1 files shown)


vuu/src/main/scala/org/finos/vuu/provider/join/JoinRelations.scala
```diff
@@ -31,6 +31,8 @@ class JoinRelations {
 
       if (leftColumn != null && leftValue != null) {
         rowJoin.putJoinKey(leftColumn, leftValue.toString)
+      } else if (ev.containsKey(leftColumn) && leftValue == null) {
+        rowJoin.deleteJoinKey(leftColumn)
```
tanjjj: after this line is added, EditableViewportWithRpcTest started to fail. Because ConstituentInstrumentPricesRpcService.editCell() does this `table.processUpdate(key, RowWithData(key, Map("id" -> key, column -> data)))`

tanjjj: ultimately we need sendToJoinSink(rowKey, rowData) to have the merged row

stonesmi (approved):

PR #2069, opened 2026-02-26
Improve removal speed on ImmutableArray

stonesmi:
Rather than immediately removing an element, soft delete it until we hit a threshold, at which we compact and do a hard delete.


stonesmi:
alas, my first draft made apply(int) become O(n), even if removal was much faster :(

PR #2172, opened 2026-06-05
#2166 add default vuu msg column to session table

### LINKED ISSUE #2166: Add default column vuu_msg to session table
Add a default column vuu_msg to session table so that we can store error messages for each row in the session table and display the errors on UI. This helps table edit mode feature.

tanjjj:
closes #2166


stonesmi (approved):

tanjjj:
@anandBalachandar @heswell FYI a column called vuuMsg is added as default column to all tables and the data type is String. Join table has "includeDefaultColumns" false so it doesn't have this column.

Issue #2169, opened 2026-06-04
Add RPC message and handler for edit mode


tanjjj:
For table edit mode, add the following:

Add BeginEditSession and EndEditSession RPC messages.
Add a parameter to indicate copying all rows/selected rows/no rows from the original table. (selected-rows, all-rows, empty-session-table).
Add a parameter for which columns to copy.
Add a parameter for which columns are editable in the session table.

In the RPC handler, when handling BeginEditSession, create a session table and copy data from original table.

Other changes related to edit mode:

Add a flag in tableDef to indicate whether the table is editable.
For each column, add "isEditable" flag. Default columns like vuuCreateTimestamp shouldn't be editable by user.

In GetTableMetaResponse, add the info of which columns are editable so this info is passed to UI.

tanjjj:
"Add a parameter for which columns are editable in the session table." - this should be in session table def

tanjjj:
Note with the current code, when copying data for selected rows, the order of rows in source table is not preserved.
We can add sorting as improvement later.

PR #2178, opened 2026-06-16
#2169 add createSessionTable and EndEditSession rpc and add isEditable to tableDef and column

tanjjj:
part 1 for #2169 

Add createSessionTable and EndEditSession rpc and CreateSessionTableRpcHandler and EndEditSessionRpcHandler

Add isEditable to TableDef. False by default. Allow true for standard table. If required in the future we can implement this for join table, sessino table etc.

Add isEditable to Column. False by default. Allow true for SimpleColumn. If required in the future we can implement this for join column, calculated column etc. For default columns, set to false by default.


vuu/src/main/scala/org/finos/vuu/api/ColumnBuilder.scala
```diff
@@ -8,61 +8,61 @@ class ColumnBuilder {
 
   val columns = new mutable.ArrayBuilder.ofRef[String]()
 
-  def addString(columnName: String): ColumnBuilder = {
```
stonesmi, line 11: can we add an overload that doesnt take in the boolean (defaulted to false) for the existing java users?

tanjjj:
@anandBalachandar @heswell FYI this PR adds the new RPC for edit mode. RpcNames.scala has the new names and EditInSessionTableRpcWSApiTest.scala has an example of the RpcParams

vuu/src/test/scala/org/finos/vuu/core/table/ColumnTest.scala
```diff
@@ -189,4 +189,25 @@ class ColumnTest extends AnyFeatureSpec with Matchers {
 
   }
 
+  Feature("Editable columns") {
+
+    Scenario("Create join columns for all columns in table def except default columns") {
```
stonesmi: copy/paste i think? can be fixed in next PR

tanjjj: oops fixed

stonesmi (approved):

PR #2185, opened 2026-06-25
Decimal pattern should accept numbers without trailing decimals e.g. "123."

jfstephe:
Added some tests too.


heswell:
well, well, I would have assumed these are not valid decimals, but google tells me they are, ylsned.

PR #2196, opened 2026-07-03
Edit Session - Allow add rows, delete rows, undo row changes

anandBalachandar:
In addition to editing cells, enabling deletions and add row(s) within EditSession

### DIFF
--- a/vuu-ui/packages/vuu-data-local/src/array-data-source/array-data-source.ts
+++ b/vuu-ui/packages/vuu-data-local/src/array-data-source/array-data-source.ts
@@ -687,7 +687,7 @@ export class ArrayDataSource
 
   protected insert = (row: VuuRowDataItemType[]) => {
     // TODO take sorting, filtering. grouping into account
-    const dataSourceRow = toDataSourceRow(this.key)(row, this.size);
+    const dataSourceRow = toDataSourceRow(this.key, this.index)(row, this.size);
     (this.#data as DataSourceRow[]).push(dataSourceRow);
     const { from, to } = this.#range;
     const [rowIdx] = dataSourceRow;
--- a/vuu-ui/packages/vuu-data-local/src/array-data-source/array-data-source.ts
+++ b/vuu-ui/packages/vuu-data-local/src/array-data-source/array-data-source.ts
@@ -13,6 +13,7 @@ import {
   WithBaseFilter,
   WithFullConfig,
   DataSourceDeleteHandler,
+  DeleteRowMode,
 } from "@vuu-ui/vuu-data-types";
 import { filterPredicate, parseFilter } from "@vuu-ui/vuu-filter-parser";
 import type {
@@ -69,6 +70,9 @@ import { Filter } from "@vuu-ui/vuu-filter-types";
 const { debug, info } = logger("ArrayDataSource");
 
 const { KEY } = metadataKeys;
+
+const VUU_MSG_SOFT_DELETED = "SOFT_DELETED";
+const VUU_MSG_COL = "vuuMsg";
 
 export interface ArrayDataSourceConstructorProps
   extends Omit<DataSourceConstructorProps, "bufferSize" | "table"> {
@@ -151,6 +155,7 @@ export class ArrayDataSource
   #range = Range(0, 0);
   #status: DataSourceStatus = "initialising";
   #title: string | undefined;
+  #markedForDeletion = new Set<string>();
 
   protected _menu: VuuMenu | undefined;
   protected selectedRows = new Set<string>();
@@ -892,8 +897,30 @@ export class ArrayDataSource
     }
   };
 
-  deleteRow: DataSourceDeleteHandler = async (key) => {
+  deleteRow: DataSourceDeleteHandler = async (key, mode: DeleteRowMode = "hard") => {
     const dataIndex = this.#data.findIndex((row) => row[KEY] === key);
+
+    if (dataIndex === -1) {
+      return "row not found";
+    }
+
+    if (mode === "soft") {
+      if (!this.#markedForDeletion.has(key)) {
+        this.#markedForDeletion.add(key);
+
+        const vuuMsgColIndex = this.#columnMap[VUU_MSG_COL];
+        if (vuuMsgColIndex !== undefined) {
+          this.#data[dataIndex][vuuMsgColIndex] = VUU_MSG_SOFT_DELETED;
+          const { from, to } = this.#range;
+          if (dataIndex >= from && dataIndex < to) {
+            this.sendRowsToClient(false, this.#data[dataIndex]);
+          }
+        }
+      }
+      return true;
+    }
+
+    this.#markedForDeletion.delete(key);
     let doomedIndex: number | undefined = undefined;
 
     if (dataIndex !== -1) {
--- a/vuu-ui/packages/vuu-data-local/src/array-data-source/array-data-source.ts
+++ b/vuu-ui/packages/vuu-data-local/src/array-data-source/array-data-source.ts
@@ -12,8 +12,6 @@ import {
   TableSchema,
   WithBaseFilter,
   WithFullConfig,
-  DataSourceDeleteHandler,
-  DeleteRowMode,
 } from "@vuu-ui/vuu-data-types";
 import { filterPredicate, parseFilter } from "@vuu-ui/vuu-filter-parser";
 import type {
@@ -70,9 +68,6 @@ import { Filter } from "@vuu-ui/vuu-filter-types";
 const { debug, info } = logger("ArrayDataSource");
 
 const { KEY } = metadataKeys;
-
-const VUU_MSG_SOFT_DELETED = "SOFT_DELETED";
-const VUU_MSG_COL = "vuuMsg";
 
 export interface ArrayDataSourceConstructorProps
   extends Omit<DataSourceConstructorProps, "bufferSize" | "table"> {
@@ -155,8 +150,6 @@ export class ArrayDataSource
   #range = Range(0, 0);
   #status: DataSourceStatus = "initialising";
   #title: string | undefined;
-  #markedForDeletion = new Set<string>();
-
   protected _menu: VuuMenu | undefined;
   protected selectedRows = new Set<string>();
   protected index = new Map<string, number>();
@@ -897,66 +890,41 @@ export class ArrayDataSource
     }
   };
 
-  deleteRow: DataSourceDeleteHandler = async (key, mode: DeleteRowMode = "hard") => {
+  protected handleDeleteFromTable = async (key: string): Promise<true | string> => {
     const dataIndex = this.#data.findIndex((row) => row[KEY] === key);
-
     if (dataIndex === -1) {
       return "row not found";
     }
-
-    if (mode === "soft") {
-      if (!this.#markedForDeletion.has(key)) {
-        this.#markedForDeletion.add(key);
-
-        const vuuMsgColIndex = this.#columnMap[VUU_MSG_COL];
-        if (vuuMsgColIndex !== undefined) {
-          this.#data[dataIndex][vuuMsgColIndex] = VUU_MSG_SOFT_DELETED;
-          const { from, to } = this.#range;
-          if (dataIndex >= from && dataIndex < to) {
-            this.sendRowsToClient(false, this.#data[dataIndex]);
-          }
-        }
-      }
-      return true;
-    }
-
-    this.#markedForDeletion.delete(key);
     let doomedIndex: number | undefined = undefined;
-
-    if (dataIndex !== -1) {
-      if (this.processedData) {
-        for (let i = 0; i < this.processedData.length; i++) {
-          if (this.processedData[i][KEY] === key) {
-            doomedIndex = i;
-          } else if (doomedIndex !== undefined) {
-            this.processedData[i][0] -= 1;
-          }
-        }
-        if (doomedIndex !== undefined) {
-          this.processedData.splice(doomedIndex, 1);
+    if (this.processedData) {
+      for (let i = 0; i < this.processedData.length; i++) {
+        if (this.processedData[i][KEY] === key) {
+          doomedIndex = i;
+        } else if (doomedIndex !== undefined) {
+          this.processedData[i][0] -= 1;
         }
       }
-
-      this.#data.splice(dataIndex, 1);
-      for (let i = dataIndex; i < this.#data.length; i++) {
-        this.#data[i][0] -= 1;
+      if (doomedIndex !== undefined) {
+        this.processedData.splice(doomedIndex, 1);
       }
-
-      this.sendSizeUpdateToClient();
-
-      const { from, to } = this.#range;
-      const deletedIndex = doomedIndex ?? dataIndex;
-      if (deletedIndex >= from && deletedIndex < to) {
-        this.#keys.reset(this.range.withBuffer);
-        this.sendRowsToClient(true);
-      }
-
-      this.emit("resize", this.size);
-
-      return true;
-    } else {
-      return "row not found";
     }
+
+    this.#data.splice(dataIndex, 1);
+    for (let i = dataIndex; i < this.#data.length; i++) {
+      this.#data[i][0] -= 1;
+    }
+
+    this.sendSizeUpdateToClient();
+
+    const { from, to } = this.#range;
+    const deletedIndex = doomedIndex ?? dataIndex;
+    if (deletedIndex >= from && deletedIndex < to) {
+      this.#keys.reset(this.range.withBuffer);
+      this.sendRowsToClient(true);
+    }
+
+    this.emit("resize", this.size);
+    return true;
   };
 
   private setRange(range: Range, forceFullRefresh = false) {
--- a/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
+++ b/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
@@ -828,21 +828,23 @@ export class VuuDataSource extends BaseDataSource implements DataSource {
     return response?.errorMessage ?? "deleteRow failed";
   }
 
+  async deleteSelectedRows(mode: DeleteRowMode = "soft"): Promise<RpcResultSuccess | RpcResultError> {
+    const rpcHost = this.#sessionDataSource ?? this;
+    const response = await rpcHost.rpcRequest?.({
+      type: "RPC_REQUEST",
+      rpcName: "deleteSelectedRows",
+      params: { mode },
+    });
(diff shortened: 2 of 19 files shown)


vuu-ui/packages/vuu-utils/src/data-editing/EditSession.tsx
```diff
@@ -26,22 +26,33 @@ type RowEditDetails = {
...
+  #deleteCount = 0;
   #invalidCount = 0;
+  #selectionCount = 0;
+  #deleteMode: DeleteRowMode;
   #sourceTableDataSource?: EditApi;
   #sessionDataSource?: EditApi;
   #inEditMode = false;
 
-  constructor(dataSource: EditApi) {
+  #handleRowSelection = (count: number) => {
```
heswell: this doesn't belong on here, its duplicating exactly what the datasSource already does.

vuu-ui/packages/vuu-utils/src/data-editing/EditSession.tsx
```diff
@@ -76,15 +87,65 @@ export class EditSession extends EventEmitter<EditSessionEvents> {
...
+      this.#deleteCount = val;
+      if (val === 0 && this.#editCount === 0) {
+        this.emit("editState", "clean");
+      } else if (oldCount === 0) {
+        this.emit("editState", "dirty");
+      }
+    }
+  }
+
+  get selectionCount() {
```
heswell: why do we need to replicate dataSource selectionCount on the editSession ?

vuu-ui/packages/vuu-utils/src/data-editing/EditSession.tsx
```diff
@@ -76,15 +87,65 @@ export class EditSession extends EventEmitter<EditSessionEvents> {
...
+    for (const key of keys) {
+      if (!this.#deletedRows.has(key)) {
+        this.#deletedRows.add(key);
+        this.deleteCount = this.#deleteCount + 1;
+        this.dataSource?.deleteRow?.(key, this.#deleteMode);
+      }
+    }
+  }
+
+  restoreRows(keys: string[]) {
```
heswell, line 140: good

vuu-ui/packages/vuu-data-types/index.d.ts
```diff
@@ -496,8 +496,10 @@ export declare type DataSourceEvents = {
   "visual-link-removed": () => void;
 };
 
+export declare type DeleteRowMode = "soft" | "hard";
```
heswell, line 507: can you add a comment explaining what hard/soft delete modes mean mean

vuu-ui/packages/vuu-utils/src/data-editing/EditButtons.tsx
```diff
@@ -4,35 +4,50 @@ import { useCallback, useMemo, useState } from "react";
...
   const [editState, setEditState] = useState<EditState>("clean");
+  const [selectionCount, setSelectionCount] = useState(0);
 
   const handleSave = useCallback(() => {
     onSave(editState === "stale");
   }, [editState, onSave]);
 
   useMemo(() => {
     editSession?.on("editState", setEditState);
+    editSession?.on("selectionCount", setSelectionCount);
```
heswell: editSession.dataSource.on(""row-selection", setSelectionCount)

vuu-ui/packages/vuu-data-local/src/array-data-source/array-data-source.ts
```diff
@@ -894,8 +899,30 @@ export class ArrayDataSource
     }
   };
 
-  deleteRow: DataSourceDeleteHandler = async (key) => {
+  deleteRow: DataSourceDeleteHandler = async (key, mode: DeleteRowMode = "hard") => {
```
heswell: this implementation needs to go in TickingArrayDataSource, where it should do an rpc request, see the implementation of editCell

vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
```diff
@@ -814,14 +815,50 @@ export class VuuDataSource extends BaseDataSource implements DataSource {
...
+      return true;
+    }
+    return response?.errorMessage ?? "deleteRow failed";
+  }
+
+  async addRow(
+    rowData: Record<string, VuuRowDataItemType> = {},
+  ): Promise<true | string> {
+    const keyColumn = this.tableSchema?.key ?? "id";
+    //Don't think we need to pass a unique id as service will generate one.
```
heswell: I agree, ui shouldn't send the key, it will always be the server where it gets set.

vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
```diff
@@ -814,14 +815,50 @@ export class VuuDataSource extends BaseDataSource implements DataSource {
...
+      rpcName: "addRow",
+      params: { key: rowKey, data: rowDataWithKey },
+    });
+    if (isRpcSuccess(response)) {
+      return true;
+    }
+    return response?.errorMessage ?? "addRow failed";
+  }
+
+  async undoRowChange(
```
heswell, line 856: is this intended to cover all possiblr cases ? editCell/deleteRow(soft) ands addRow

vuu-ui/packages/vuu-data-test/src/core/module/VuuModule.ts
```diff
@@ -405,32 +406,69 @@ export abstract class VuuModule<T extends string = string>
...
+        }
+        return { type: "SUCCESS_RESULT", data: undefined };
       } else {
         const { dataSource } = this.getSubscriptionByViewport(viewPortId);
         if (dataSource.table) {
           const table = this.tables[dataSource.table.table as T];
           if (table) {
-            table.delete(key as string);
+            if (mode === "soft") {
+              table.update(key, "vuuMsg", "SOFT_DELETED");
```
heswell: will we actually overload vuuMsg for this purpose ? It might be better to add a separate boolean col explicitly to track soft deletes

vuu-ui/packages/vuu-data-test/src/core/module/VuuModule.ts
```diff
@@ -405,32 +406,69 @@ export abstract class VuuModule<T extends string = string>
...
+    return { type: "ERROR_RESULT", errorMessage: "something went wrong" };
+  };
+
+  private undoRowChange: ServiceHandler = async (rpcRequest) => {
+    if (isUndoRowChangeRpcRequest(rpcRequest)) {
+      const { viewPortId } = rpcRequest.context;
+      const { key } = rpcRequest.params;
+      const sessionTable = this.getSessionTable(viewPortId, false);
+      if (sessionTable && isProxySessionTable(sessionTable)) {
+        const updates = sessionTable.getSessionUpdates();
```
heswell, line 503: I need to update this code. The 'proxysession table' is going away. It was an experiment at an efficient representation of a session table that tried to avoid copying data. If falls down though when the data is ticking in any way and would deviate too much from the server implementation. 
And in reality copying even a few hundred thousand rows takes hardly any time.

vuu-ui/packages/vuu-data-test/src/SessionTable.ts
```diff
@@ -23,6 +23,8 @@ export const isProxySessionTable = (table: any): table is SessionTable =>
 // How can we ignore these ?
 export const SessionTable = (table: Table, sessionId: string): SessionTable => {
   const updates = new Map<string, RowUpdates>();
+  /** Rows added in this session. */
```
heswell, line 26: sorry my bad. This is going to be removed, I just haven't found the time to do it. We need to revert to a very simple createSessionTable that just copies across all rows.

lets discuss

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
```diff
@@ -168,10 +185,36 @@ export class TickingArrayDataSource extends ArrayDataSource {
     return this.#getVisualLinks?.(this.table.table);
   }
 
-  getSelectedRowIds() {
+  getSelectedRowIds(): string[] {
+    if (this.#sessionDataSource) {
+      // Merge base and session selections.
```
heswell: This doesn't feel right. I think I understand the reason, but merging the selections from two tables doesn't feel like the right solution . One thing always to bear in mind when changing an API - even though a change might work nicely for a current use case, other users in the future will use this for other use cases and I'm not sure this would be what users would expect.

Lets discuss

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
```diff
@@ -168,10 +185,36 @@ export class TickingArrayDataSource extends ArrayDataSource {
...
+    }
     return Array.from(this.selectedRows);
   }
 
+  /**
+   * Suppress row updates from the main data source during an edit session
+   * to prevent overwriting the session view.
+   */
+  sendRowsToClient(forceFullRefresh = false, row?: DataSourceRow) {
+    if (this.#sessionDataSource) {
```
heswell, line 197: I don't think this should be necessary, it implies other checks are not working as expected, which we should get to the bottom of

vuu-ui/packages/vuu-utils/src/data-editing/README.md
```diff
@@ -0,0 +1,286 @@
...
+  → editSession.end(save?, force?)    // commits or discards, clears all state
+```
+
+`begin()` calls `dataSource.beginEditSession()` which creates a **session datasource** (a proxy/mirror of the source table). All subsequent cell edits and row operations operate against this session datasource, leaving the source table untouched until `end(true)` is called.
+
+### Row Operations
+
+| Method | Description |
+|---|---|
+| `editSession.addRows(count, rowData?)` | Adds `count` blank rows via the source-table datasource `addRow` RPC |
```
heswell, line 49: whats the use case for adding n blank rows ?

vuu-ui/packages/vuu-utils/src/data-editing/README.md
```diff
@@ -0,0 +1,286 @@
...
+| `onDelete` | Deletes selected rows via `editSession.deleteRows()` |
+| `onAddRows` | Adds rows via `editSession.addRows(addRowsCount)` |
+| `onUndoRowChange` | `(key) =>` `editSession.undoRowChange(key)` |
+| `onSelectionChange` | Keeps `selectedKeysRef` and `hasSelection` up to date |
+
+---
+
+## EditButtons — UI Component
+
+Renders the action bar for an edit session. Consumes `EditSession` directly for state, takes handler callbacks from `useEditableTable`.
```
heswell: useEditableTable can be used to manage the cancel/save operations but its use is not a hard requiremenbt. However that can be worded

vuu-ui/packages/vuu-utils/src/data-editing/useEditableTable.ts
```diff
@@ -90,22 +94,43 @@ export const useEditableTable = ({
     [dataSource, editSession, onSave],
   );
 
+  const handleDelete = useCallback(() => {
+    if (selectedKeysRef.current.length > 0) {
+      editSession.deleteRows(selectedKeysRef.current);
```
heswell: remember selection state resides on the server. Ui shouln't need to send selected row details to server.We probably need a 'deleteSelectedRows' rpcName or a 'selected' parameter, then we can leave it to the server to delete rows based on the current selection

vuu-ui/packages/vuu-utils/src/data-editing/useEditableTable.ts
```diff
@@ -90,22 +94,43 @@ export const useEditableTable = ({
...
+  const handleAddRows = useCallback(() => {
+    editSession.addRows(addRowsCount);
+  }, [addRowsCount, editSession]);
+
+  const handleUndoRowChange = useCallback(
+    (key: string) => void editSession.undoRowChange(key),
+    [editSession],
+  );
+
+  const handleSelectionChange = useCallback<SelectionChangeHandler>(
```
heswell: I don't think we will need this (see other comments)

vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
anandBalachandar: yes, this is required where we want to display an undo option at row level - which will then reinstate the row from the base table.

vuu-ui/packages/vuu-data-test/src/core/module/VuuModule.ts
anandBalachandar: This is basically to simulate a "server" behavior where we want to mark the rows for deletion but not remove the row from the table. The server logic would update a specific column based on the table schema/columns that are being used. This only affects the local datasource and we can check this with the inline edit (instruments) example.

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
anandBalachandar: During an inline edit session, newly added rows exist only in the session datasource - they are inserted via sessionTable.insert() so the source table doesn't know about them. Their selection state therefore only lives in #sessionDataSource.selectedRows and never in this.selectedRows.

Without the merge, selecting a freshly added row and then clicking Delete would read an empty selection from the source datasource and silently doesnt do anything...

Trying to replicate remote server behaviour...

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
anandBalachandar: The source table continues to receive live ticking updates while the user is editing. If super.sendRowsToClient were allowed to run, it would push the original (unedited) source rows directly to the client, overwriting whatever edited values the session datasource has already sent. The user would see their edits disappear. 

Also, The session datasource sends its own viewport-update messages through handleSessionMessage - clientCallback.  If the source datasource also called super.sendRowsToClient, the client would receive two streams of row updates for the same viewport - one with session edits, one without?

vuu-ui/packages/vuu-utils/src/data-editing/README.md
anandBalachandar: To cover the use case for adding a fixed set of empty rows on "Add Rows" in the table..
This might change.

vuu-ui/packages/vuu-data-test/src/core/module/VuuModule.ts
anandBalachandar: Injecting the message column in VuuModule now instead of hard wiring it.

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
heswell: The reason I don't understand why this is required is that we handle updates to the underlying table with the updateRowWithSessionCheck handler. This ignores updates when we have a sessionDataSource in play. How are we ending up in this code ?

vuu-ui/packages/vuu-data-test/test/TickingArrayDataSource.EditApi.test.ts
```diff
@@ -0,0 +1,392 @@
+import { describe, expect, it, vi } from "vitest";
```
heswell, line 1: no objection to unit tests but with so much mocking going on, I think there is more value in the playwright component tests that are testing the data source in actual operation. I haven't created nearluy ewnough tests yet in Table-editing.playwright.test

vuu-ui/packages/vuu-data-test/src/simul/simul-schemas.ts
```diff
@@ -29,6 +33,7 @@ export const schemas: Readonly<Record<SimulTableName, Readonly<TableSchema>>> =
         { name: "lotSize", serverDataType: "int" },
         { name: "ric", serverDataType: "string" },
         ...VUU_TIMESTAMP_COLUMNS,
+        ...VUU_MSG_COLUMN,
```
heswell: I don't think vuuMsg should be on the source table schema, its a session table thing only isn't it ?

Seems I'm mistaken, its going to be on all columns. Not sure that seems right to me, but not a big deal

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
heswell: got it. I think we should return either the selection from the session table OR , if we have no edit session, the standard selection.
Can't see how a merged set of selected rows from two tables would be needed.

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
anandBalachandar: updateRowWithSessionCheck already blocks individual row updates arriving  via the table's update event.  
sendRowsToClient can also be reached through other paths that bypass that handler? setRange/config changes?

vuu-ui/packages/vuu-data-test/test/TickingArrayDataSource.EditApi.test.ts
anandBalachandar: Added some playwright component tests

vuu-ui/packages/vuu-data-test/src/TickingArrayDataSource.ts
heswell: Lets leave this check in, then, but lets console.warn a message. We should then examine all the places where this could be invoked during an edit session, so we can be sure we're not covering up an issue here that requires further attention

vuu-ui/showcase/src/examples/Table/Editing.examples.tsx
```diff
@@ -66,6 +66,7 @@ const InstrumentColumns: ColumnDescriptor[] = [
   { name: "ric", serverDataType: "string", width: 75 },
   { name: "vuuCreatedTimestamp", serverDataType: "epochtimestamp", width: 160 },
   { name: "vuuUpdatedTimestamp", serverDataType: "epochtimestamp", width: 160 },
+  { name: "vuuMsg", serverDataType: "string", width: 120 },
```
heswell, line 69: Need to make sure this column is not rendered as editable. Change needed in code below. This is why playwright test breaks . Its testing that only editable cells are included in focus traversal. 

```
                  : col.name === "isin" ||
                      col.name === "vuuCreatedTimestamp" ||
                      col.name === "vuuUpdatedTimestamp"
                    ? col
                    : { ...col, editable: true },
```

given the number of columns, more succinct to use regex

```
col.name.match(/^(isin|vuuCreatedTimestamp|vuuUpdatedTimestamp|vuuMsg)$/)
```

or test against an array of names - whatever.  Server is going to give us list of the editable column sin TABLE_METADATA, we will incorporate that into the table schema when its ready

PR #2206, opened 2026-07-13
#2169 copy data to session table when handling CreateSessionTableRpc

tanjjj:
Part 2 of #2169 

Changes:
- add editable columns to GetTableMetaResponse
- add VuuRpcOptions and max copy size
- implement CreateSessionTableRpcHandler to copy data from source table to session table. If rpc params specify the list of columns to copy from, then data is copied only for these columns. Otherwise if "*" is used, all columns are copied.

### DIFF
--- a/vuu/src/main/scala/org/finos/vuu/net/Messages.scala
+++ b/vuu/src/main/scala/org/finos/vuu/net/Messages.scala
@@ -44,7 +44,7 @@ case class GetTableListResponse(tables: Array[ViewPortTable]) extends MessageBod
 
 case class GetTableMetaRequest(table: ViewPortTable) extends MessageBody
 
-case class GetTableMetaResponse(table: ViewPortTable, columns: Array[String], dataTypes: Array[String], key: String) extends MessageBody
+case class GetTableMetaResponse(table: ViewPortTable, columns: Array[String], dataTypes: Array[String], key: String, editableColumns: Array[String]) extends MessageBody
 
 case class GetViewPortMenusRequest(vpId: String) extends MessageBody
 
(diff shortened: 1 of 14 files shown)


vuu/src/main/scala/org/finos/vuu/net/rpc/CreateSessionTableRpcHandler.scala
```diff
@@ -14,39 +11,52 @@ class CreateSessionTableRpcHandler(using val tableContainer: TableContainer) ext
...
+          }
+          i += 1
+        }
       case Selected =>
-      //TODO #2169 copy selected rows
-      // sort keys? val keys = aSort.doSort(table.sourceTable, filteredKeys, vpColumns)
+        val vp = params.viewPort
+        val vpKeys = vp.getKeys
+        val selection = vp.getSelection
+        val vpColumns = ViewPortColumnCreator.create(params.viewPort.table.asTable, columnsToCopy.toList)
```
stonesmi: this needs to check the max size, as the user could have selected all

vuu/src/main/scala/org/finos/vuu/net/Messages.scala
```diff
@@ -44,7 +44,7 @@ case class GetTableListResponse(tables: Array[ViewPortTable]) extends MessageBod
 
 case class GetTableMetaRequest(table: ViewPortTable) extends MessageBody
 
-case class GetTableMetaResponse(table: ViewPortTable, columns: Array[String], dataTypes: Array[String], key: String) extends MessageBody
+case class GetTableMetaResponse(table: ViewPortTable, columns: Array[String], dataTypes: Array[String], key: String, editableColumns: Array[String]) extends MessageBody
```
stonesmi, line 47: will the ui ignore the extra field without changes? @heswell

vuu/src/main/scala/org/finos/vuu/net/rpc/CreateSessionTableRpcHandler.scala
tanjjj: updated

vuu/src/main/scala/org/finos/vuu/net/Messages.scala
heswell: yes, but we'll add it to the tableSchema struct that we build on the client

stonesmi (approved):

PR #2213, opened 2026-07-16
Add EpochTimestampNano data type

stonesmi:
Trading systems are generally required to support Nanosecond level precision on timestamps. 

I've added support for this alongside our existing millisecond level type.

I also refactored the JSON serializer for RowData to make it allocate less.


stonesmi:
@heswell, we will send the Nanosecond precision timestamps as Strings, similar to ScaledDecimal and Long.

tanjjj (approved):

PR #2219, opened 2026-07-18
Add configurable limits for maximum VP scroll depth and range width

stonesmi:
We should prevent Vuu clients from requesting too much data, particularly on virtualised grids with large amounts of data.

To this end, I have added two new fields on TableMetadata to let the Client know what is and isnt acceptable. These values are configurable on the server side.

```
maxRangeEnd: Int
maxRangeWidth: Int
```

Should the UI violate these values, they will receive a ChangeViewPortRangeReject message, which is new:

`case class ChangeViewPortRangeReject(viewPortId: String, msg: String) extends MessageBody`

The server will internally set the Range to 0 -> 0, until a valid range is requested.


stonesmi:
@heswell this is ready for you to test against.

tanjjj (approved):

PR #2222, opened 2026-07-21
feat: add createSessionTable RPC with CopyOption API

anandBalachandar:
- Introduce CopyOption = "All" | "Empty" | "Selected" for standalone session tables
- Remove CopyOption from EditSessionMode union; deprecate EditSessionMode for new code
- Fix useEditableTable: subscribe row-selection on active datasource (sessionDataSource ?? dataSource); reset sessionDataSource/selectionCount on session end
- Add editSessionMode prop to EditableInstrumentsTemplate with loading state for async createSessionTable; add CreateSessionTableInstruments showcase example
- Added unit tests (createSessionTable), Playwright CT tests

### DIFF
--- a/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
+++ b/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
@@ -680,18 +680,7 @@ export class VuuDataSource extends BaseDataSource implements DataSourceBase {
     return Promise.reject<T>();
   }
 
-  createSessionDataSource(sessionTable: VuuTable) {
-    //TODO filters, sort etc
-    const columns = this.#sessionTableMessageColumn
-      ? this.columns.concat(this.#sessionTableMessageColumn)
-      : this.columns;
-    return new VuuDataSource({
-      columns: columns,
-      table: sessionTable,
-    });
-  }
-
-  async createSessionTable(
+  async createSessionDataSource(
     copyOption: CopyOption,
   ): Promise<VuuDataSource | undefined> {
     const rpcResponse = await this?.rpcRequest?.({
@@ -708,7 +697,7 @@ export class VuuDataSource extends BaseDataSource implements DataSourceBase {
       });
     } else {
       throw Error(
-        `[VuuDataSource] createSessionTable ${rpcResponse?.errorMessage}`,
+        `[VuuDataSource] createSessionDataSource ${rpcResponse?.errorMessage}`,
       );
     }
   }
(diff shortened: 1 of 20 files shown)


vuu-ui/packages/vuu-data-test/src/core/module/VuuModule.ts
```diff
@@ -642,13 +642,58 @@ export abstract class VuuModule<T extends string = string>
     }
   };
 
+  /**
+   * Creates a standalone session table without starting an inline edit session.
```
heswell: minor wording quibble. There is no such thing as an 'inline edit session' from the point of view of the service.

vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
```diff
@@ -691,12 +691,34 @@ export class VuuDataSource extends BaseDataSource implements DataSourceBase {
     });
   }
 
+  async createSessionTable(
```
heswell: we already have `createSessionDataSource`, which was added for exactly this purpose. You should extend the existing method. Nothing else uses it yet, it was added in anticipation of upload functionality.

`createSessionDataSource` is a more accurate name for what the method does, creating a session table is a side effect.

vuu-ui/packages/vuu-table-extras/test/csv-upload/useCsvUpload.test.tsx
```diff
@@ -46,7 +46,7 @@ const makeDataSource = (overrides: Partial<DataSource> = {}): DataSource =>
     table: vuuTable,
```
heswell, line 46: can we not use playwright component tests for this component ?

vuu-ui/packages/vuu-utils/src/data-editing/useEditableTable.ts
```diff
@@ -107,14 +111,17 @@ export const useEditableTable = ({
...
+    const activeDataSource = sessionDataSource ?? dataSource;
+    activeDataSource.on("row-selection", setSelectionCount);
+    return () => activeDataSource.removeListener("row-selection", setSelectionCount);
+  }, [dataSource, sessionDataSource]);
 
   useMemo(async () => {
     if (isEditMode) {
       try {
-        const sessionDataSource = await editSession.begin(editSessionMode);
+        const sessionDataSource = isCopyOption(editSessionMode)
```
heswell, line 122: I noticed this change only while debugging. Not sure what the intention was but it hasn;'t changed the original code at all

anandBalachandar: TS cant resolve which overload to use and errors on both and applying await to each branch individually rather than to the whole expression lets TS narrow the type per branch satisfying the correct overload in each case..

PR #2223, opened 2026-07-22
Extend toast notification functions

atkiss:
- implement hide notification for toast
- add link support to toast notification
- add optional custom class parameter for toast notification
- add no-wrap for toast content
- update showcase to demo latest changes
- add notification pw test to cover the latest changes

### DIFF
--- a/vuu-ui/packages/vuu-notifications/src/NotificationsCenter.tsx
+++ b/vuu-ui/packages/vuu-notifications/src/NotificationsCenter.tsx
@@ -51,7 +51,7 @@ const RuntimeToast = (
   return {
     ...toast,
     hidden: slidesIn,
-    id: getUniqueId(),
+    id: toast.id ?? getUniqueId(),
     left: document.body.clientWidth,
     opacity: slidesIn ? undefined : 0,
     size: ZeroSize,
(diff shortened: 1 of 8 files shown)


vuu-ui/packages/vuu-notifications/src/NotificationsCenter.tsx
```diff
@@ -51,7 +51,7 @@ const RuntimeToast = (
   return {
     ...toast,
     hidden: slidesIn,
-    id: getUniqueId(),
+    id: toast.id ?? getUniqueId(),
```
heswell: are we trusting clients here to give us unique ids ? isn't this a bit dangerous. Suppose they give us dupe id values

vuu-ui/packages/vuu-notifications/src/ToastNotification.tsx
```diff
@@ -8,6 +8,11 @@ import type { ToastNotificationDescriptor } from "./NotificationsContext";
 
 import toastNotificationCss from "./ToastNotification.css";
 
+import React from "react";
+
+import { Link, useInRouterContext } from "react-router-dom";
```
heswell: this is tying our Toast implementation to react router.
We can't have react-router as a dependency of Toast,

vuu-ui/packages/vuu-notifications/src/NotificationsCenter.tsx
atkiss: Removed. The only thing while I added this because this way the renderPostRefresh would give the same id to the notification but as I checked it does not really matter if it receives a new id.

vuu-ui/packages/vuu-notifications/src/ToastNotification.tsx
atkiss: Link handling is removed completly as we discussed.

PR #2228, opened 2026-07-23
feat(table): Add prop isRowSelectable

anandBalachandar:
feat(table): Add prop isRowSelectable

- isRowSelectable: (dataRow: DataRow) => boolean on TableProps.
- Non-selectable rows are skipped in useSelection (all selection models — checkbox, extended, single, block). 
- They receive a vuuTableRow-noSelect CSS class which styles the checkbox as visually disabled. 
- The checkbox checked state is preserved for non-selectable rows even when isSelected is cleared by a subsequent selection.
- Add Playwright CT test for soft-delete checkbox persistence across row selection
- Document in README


vuu-ui/packages/vuu-table/src/cell-renderers/checkbox-row-selector/CheckboxRowSelectorCell.tsx
```diff
@@ -21,7 +22,15 @@ export const CheckboxRowSelectorCell: React.FC<TableCellRendererProps> = ({
     window: targetWindow,
   });
 
-  const isChecked = !!dataRow.isSelected;
+  const { checkboxRowLevelProps } =
```
heswell: quite a lot going on here for a single statement. might be better broken down a little.

vuu-ui/packages/vuu-table/src/Table.tsx
```diff
@@ -279,6 +281,11 @@ export interface TableProps
    * `checkbox` same behaviour as extended, with checkbox column for selection
    */
   selectionModel?: TableSelectionModel;
+  /**
+   * When provided, the row checkbox is rendered as disabled for any row
+   * for which this predicate returns true.
+   */
+  checkboxRowLevelProps?: (dataRow: DataRow) => CheckboxRowLevelProps;
```
heswell: props should be in alphabetic order (with capitalised props forst)

vuu-ui/packages/vuu-table-types/index.d.ts
```diff
@@ -220,7 +220,19 @@ export interface ColumnTypeRendering {
   // specific to Background renderer
   flashStyle?: "bg-only" | "arrow-bg" | "arrow";
   name: string;
+  /** Arbitrary props passed to the registered cell renderer component. */
+  componentProps?: Record<string, unknown>;
```
heswell: this cannot be unknown. It can only be a simple type - string, number or boolean as it mist be serializeable.

heswell (commented):
We should park this. The problem is is intended to solve is better addressed with the solution we discussed offline.

vuu-ui/packages/vuu-table-types/index.d.ts
```diff
@@ -405,6 +407,11 @@ export interface RuntimeColumnDescriptor extends ColumnDescriptor {
   flex?: number;
   heading?: [...string[]];
   isGroup?: boolean;
+  /**
+   * When provided on the checkbox column, rows returning false
+   * are always shown as checked (pending deletion) regardless of isSelected.
+   */
+  isRowSelectable?: (dataRow: DataRow) => boolean;
```
heswell: We have this on tableProps, I don't think we need it on a column as well

vuu-ui/packages/vuu-table/src/cell-renderers/checkbox-row-selector/CheckboxRowSelectorCell.tsx
```diff
@@ -21,7 +22,10 @@ export const CheckboxRowSelectorCell: React.FC<TableCellRendererProps> = ({
     window: targetWindow,
   });
 
-  const isChecked = !!dataRow.isSelected;
+  const { isRowSelectable } = column;
```
heswell: I don't think this is a general rule. It might be the case that a 'soft deleted' row could be assumed to be checked, if selection is being used for delete.
There are definitely other use cases though where we might want to make a row non-selectable based on data, where we would not want the checkbox to appear selected.

we can effect this locally be taking advantage of the noSelect className. eg.if we want the unselectable checkboxes to appear dimmed but checkin in a specific scenarion, we can use local css and the noSelect class to achieve this.

vuu-ui/packages/vuu-table/src/useTableModel.ts
```diff
@@ -308,16 +314,16 @@ function init(
...
-          : CheckboxColumnDescriptor(selectionModel, checkboxColumnWidth),
-        1,
-      ),
+    const checkboxCol = toRuntimeColumnDescriptor(
+      somePinnedLeft
+        ? PinnedCheckboxColumnDescriptor(selectionModel, checkboxColumnWidth)
+        : CheckboxColumnDescriptor(selectionModel, checkboxColumnWidth),
+      1,
     );
+    if (isRowSelectable) {
```
heswell: not needed

vuu-ui/packages/vuu-table/src/useSelection.ts
heswell, line 1: This is all good but you're dodging the trickiest part which is the block selection logic (see selection-utils)

If user selects a row, then shift selects to select a block, we must not include any rows that are not selectable.
This will mean changing the single block request to one or more block requests that omit the non-selectable rows

anandBalachandar: Fair point 😄. I wasn't trying to dodge that bit - I just wanted to sanity check the overall approach first before tackling the trickier selection logic. I'll update it to split the block request around any non-selectable rows.

vuu-ui/packages/vuu-table/src/useSelection.ts
```diff
@@ -118,7 +130,22 @@ export const useSelection = ({
         newRowIdentifier,
         rangeSelect,
       );
-
+      if (rangeSelect && isRowSelectable && activeRowKey && toRowKey) {
+        // Split the range into sub-ranges, skipping non-selectable rows
+        const minIdx = Math.min(rowIdx, activeRowKey.rowIdx);
+        const maxIdx = Math.max(rowIdx, activeRowKey.rowIdx);
+        const rowsInRange = dataRowsRef.current
```
anandBalachandar, line 137: dataRowsRef.current holds the render buffer - rows currently in the viewport + a small overrender buffer. If the shift-click range spans rows that have been virtualised out (scrolled far away), those rows are absent from the array. splitSelectableRanges will see a gap in the index sequence and treat it as if there are no non-selectable rows there - it produces a single sub-range across the gap rather than correctly splitting around any non-selectable rows that happen to sit outside the buffer..

anandBalachandar: Still looking into this..

anandBalachandar: Do we need this implementation (in datasource) to address this? 
getRowsInRange(fromIndex: number, toIndex: number): DataRow[]

heswell: This is a very good point and one Ive been thinking about. 
The _only_ way this can be fully implemented is with server support. We would need to include a filter with the SELECT_ROW_RANGE to identify rows to be excluded from the selection.

I think we should go with what we have and will have to live with the fact that (in our current use case) soff deleted rows _will_ be selected if they are somewhere deep in. a serverside selection block. Server will just have to ignore them if they are submitted for deletion twice.

vuu-ui/packages/vuu-utils/src/selection-utils.ts
```diff
@@ -43,3 +43,43 @@ export const selectItem = (
...
+  rows: DataRow[],
+  isRowSelectable: (row: DataRow) => boolean,
+  preserveExistingSelection: boolean,
+): Omit<SelectRequest, "vpId">[] => {
+  const requests: Omit<SelectRequest, "vpId">[] = [];
+  let fromKey: string | undefined;
+  let toKey: string | undefined;
+
+  for (const row of rows) {
+    if (isRowSelectable(row)) {
```
heswell, line 61: strictly , we should probably check for a range of 1 and return that as a SELECT_ROW rather than SELECT_ROW_RANGE, but this probably works anyway, so might as well keep the logic simpler

PR #2234, opened 2026-07-24
#2230 Add Column combine utility method

### LINKED ISSUE #2230: Add a generic combine method to toolbox
Use case: combine 2 arrays of columns when creating table def especially for join table, e.g. combine Column[] with another Column[]. Scala has ++, add a generic method for java, i.e. 
T[] combine( T[] left, T[]right)

tanjjj:
closes #2230


toolbox/src/main/java/org/finos/vuu/toolbox/Utils.java
```diff
@@ -0,0 +1,11 @@
+package org.finos.vuu.toolbox;
+
+import java.util.Arrays;
+
+public class Utils {
```
stonesmi: can we call it collection utils and write it in scala?

tanjjj: scala has ++ for arrays so doesn't really need this method? I'm mainly trying to make combining columns easier in java. Another option is putting this method in Columns which has other methods like allFrom(table: TableDef): Array[Column]

stonesmi: Ah I see! My only worry with having both java and scala code in the same module is compiler ordering tends to upset intellij.

Columns.combine might be safer, unless you can write a scala class to combine two java arrays in toolbox

tanjjj: it's moved to Columns now

stonesmi (approved):

PR #2235, opened 2026-07-27
#2227 Added generic notifications module

### LINKED ISSUE #2227: Create a generic notifications module for toasts and banners
### Description of Problem:

We have a number of use cases for pushing notifications to the front end, such as Toasts or Banners and would like to standardise how to do that.

### Potential Solutions:

Create a set of generic Notifications columns, that can be imported into Vuu applications. The applications will be responsible by creating the Provider and providing any additional columns.

These can then be used when creating a tableDef, by using a helper method similar to Columns.allFrom

We should then integrate this module into the Vuu Sample App.

Columns:

* id => String
* type => String
* expiryTime => EpochTimestamp
* title => String
* message => String
* level => String

MarioHabor:
closes #2227


vuu/src/main/scala/org/finos/vuu/core/module/notifications/NotificationModule.scala
```diff
@@ -0,0 +1,35 @@
...
+import org.finos.vuu.core.AbstractVuuServer
+import org.finos.vuu.core.module.{DefaultModule, ModuleFactory, TableDefContainer, ViewServerModule}
+import org.finos.vuu.core.table.DataTable
+import org.finos.vuu.provider.Provider
+
+object NotificationModule extends DefaultModule {
+  final val NAME = "NOTIFICATIONS"
+  final val TABLE_NAME = "notifications"
+
+  def apply(providerFunc: (DataTable, AbstractVuuServer) => Provider, additionalColumns: String*)(implicit clock: Clock, lifecycle: LifecycleContainer, tableDefContainer: TableDefContainer): ViewServerModule = {
```
tanjjj: For this notification table, we need the option to pass in a permission filter so that the user only see notifications they are allowed to see. And in order to do permission check, we need a target/audience column in this table.

tanjjj: And do we need a column for who dismissed the notification? And also a rpc handler to handle dismiss

heswell:
It would be fun to be able to test it from the UI.
Supposing we added a Service to the Module that exposed an RPC call. Client makes an rpc request with some notification details and a notification is sent to all logged in users ? 
@stonesmi wdyt ?

example/notifications/src/main/scala/org/finos/vuu/example/notifications/module/SimulatedNotificationsModule.scala
```diff
@@ -0,0 +1,27 @@
...
+
+  def apply()(implicit clock: Clock, lifecycle: LifecycleContainer, tableDefContainer: TableDefContainer): ViewServerModule = {
+    // Example permission function: only show notifications where audience == "all" OR audience == currentUser
+    val permissionFunc = (viewPort: org.finos.vuu.viewport.ViewPort, _: org.finos.vuu.core.table.TableContainer) => {
+      val currentUser = viewPort.user.name
+      // We can use PermissionFilter's Set-based Contains matching on the 'audience' column
+      PermissionFilter("audience", Set("all", currentUser, "admin", "trader"))
+    }
+
+    NotificationModule(
```
stonesmi, line 49: can we add a WebSocket test that connects to this table using a viewport and asserts that is isnt empty?

also can we add an example rpc service that does the dismissal and sets the username please?

this also gives Steve the ability to connect to it using the sample client and play around

vuu/src/main/scala/org/finos/vuu/core/module/notifications/NotificationsSchema.scala
```diff
@@ -0,0 +1,24 @@
...
+object NotificationsSchema {
+  val Id = "id".string()
+  val Type = "type".string()
+  val ExpiryTime = "expiryTime".epochTimestamp()
+  val Title = "title".string()
+  val Message = "message".string()
+  val Level = "level".string()
+  val Audience = "audience".string()
+
+  private final val GenericColumns: Array[String] = Array(Id, Type, ExpiryTime, Title, Message, Level, Audience)
```
stonesmi: NIT: private variables should be lower camel case

stonesmi (approved):

chrisjstevo:
Can we consider how we'd collapse alerts also?

If we have 1000 alerts generated for orders cancelled, we probably want to allow a user to see the first alert and then say "you have 999 more like this"

Perhaps this is a "collapse Id" or something similar in the backend?

Beyond 20-30 alerts the usefulness drops a lot.

PR #2259, opened 2026-08-04
fix(editing): correct session lifecycle and deleteCount state management

anandBalachandar:
- Reset deleteCount on cancel, save and mode-exit
- Fix else branch to call end()
- Log error in handleCancel catch

### DIFF
--- a/vuu-ui/packages/vuu-utils/src/data-editing/useEditableTable.ts
+++ b/vuu-ui/packages/vuu-utils/src/data-editing/useEditableTable.ts
@@ -143,10 +143,12 @@ export const useEditableTable = ({
         onCancel();
       }
     } else {
-      await editSession.end();
-      setSessionDataSource(undefined);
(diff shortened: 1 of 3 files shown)


vuu-ui/packages/vuu-utils/src/data-editing/useEditableTable.ts
```diff
@@ -140,10 +142,11 @@ export const useEditableTable = ({
         console.error(`[useEditableTable] begin edit session failed`, e);
         onCancel();
       }
-    } else if (editSession.inEditMode) {
+    } else {
       await editSession.end();
```
heswell: why remove this ? If the session is not in edit mode then  none of  these should be necessary.

If these need clearing and yet session is not in edit mode, something has gone wrong with logic  and we should get to the bottom of why.

This would fire on every render if anyone passed in an unstable onCancel

PR #2281, opened 2026-08-11
First version of ItemPicker as a controlled component

Louisianna1:
Addition of ItemPicker to the vuu-ui-controls package; a generic controlled component that displays items in a vertical 'picker' that are available to search, select, reorder or deselect.
Fast follow changes planned:
- Refactor of existing ColumnPicker component in vuu-table-extras package to use the ItemPicker internally (with generic logic for update of the selected 'items' and search pattern application to be removed from ColumnModel).

Future potential enhancements:
- Support to configure selection mode (single or multiple) as well as maximum permitted number of selections in the latter case.
- Automated sorting for items in the selected list using various algorithms, for example alphabetically ascending or descending, custom.
- Display of items by functional 'grouping'.

### DIFF
--- a/vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
+++ b/vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
@@ -1,14 +1,4 @@
 import {
-  ForwardedRef,
-  forwardRef,
-  RefCallback,
-  HTMLAttributes,
-  MouseEventHandler,
-  useCallback,
-  useMemo,
-  useRef,
-} from "react";
-import {
   Button,
   Input,
   ListBox,
@@ -19,15 +9,33 @@ import {
 import { useComponentCssInjection } from "@salt-ds/styles";
 import { useWindow } from "@salt-ds/window";
 import { applyHighlighting } from "@vuu-ui/vuu-table";
-import { Icon, IconButton } from "@vuu-ui/vuu-ui-controls";
-import { DragDropProvider, useSortable } from "@vuu-ui/vuu-utils";
 import {
-  ItemDescriptor,
+  capitalize,
+  DragDropProvider,
+  ItemTypeName,
+  pluralForm,
+  singularForm,
+  useSortable,
+} from "@vuu-ui/vuu-utils";
+import cx from "clsx";
+import {
+  ForwardedRef,
+  forwardRef,
+  HTMLAttributes,
+  MouseEventHandler,
+  RefCallback,
+  useCallback,
+  useMemo,
+  useRef,
+} from "react";
+import { Icon, IconButton } from "../icon-button";
+import {
   CreateCustomItemProps,
+  getItemLabel,
+  ItemDescriptor,
   ItemPickerHookProps,
   useItemPicker,
 } from "./useItemPicker";
-import cx from "clsx";
 
 import itemPickerCss from "./ItemPicker.css";
 
@@ -39,16 +47,11 @@ export interface ItemPickerProps
   extends ItemPickerHookProps,
     HTMLAttributes<HTMLDivElement>,
     Pick<ListBoxProps<ItemDescriptor>, "selected" | "onSelectionChange"> {
-  itemTypeSingular: string;
+  itemTypeName: ItemTypeName;
   createCustomItemProps?: CreateCustomItemProps;
 }
 
-const getItemLabel = (item: ItemDescriptor) => {
-  if (item.label) return item.label;
-  else return item.name;
-};
-
-const searchIcon = <span data-icon="search" />;
+const searchIcon = <Icon name="search" />;
 const NO_SELECTION: ItemDescriptor[] = [] as const;
 
 const useSorting = (id: string, index: number) => {
@@ -128,6 +131,7 @@ const AvailableListItem = ({
   item,
   onAdd,
   searchPattern = "",
+  disabled,
   ...optionProps
 }: OptionProps & {
   index: number;
@@ -143,6 +147,7 @@ const AvailableListItem = ({
       {...optionProps}
       className={cx(classNameProp, classBaseListItem)}
       data-name={item.name}
+      disabled={disabled}
     >
       <span className={`${classBase}-text`}>{valueWithHighlighting}</span>
       <IconButton
@@ -152,6 +157,7 @@ const AvailableListItem = ({
         icon="plus"
         onClick={onAdd}
         size={16}
+        disabled={disabled}
       />
     </Option>
   );
@@ -164,12 +170,11 @@ const AvailableListItem = ({
 export const ItemPicker = forwardRef(function ItemPicker(
   {
     className,
-    itemTypeSingular,
+    itemTypeName,
     allItems,
     selectedItems,
-    searchPattern,
+    maxSelections,
     onSelectedItemsChange,
-    onSearchPatternChange,
     onSelectionChange,
     selected = NO_SELECTION,
     createCustomItemProps,
@@ -191,6 +196,7 @@ export const ItemPicker = forwardRef(function ItemPicker(
   }, []);
 
   const {
+    selectedItemsCount,
     selectedItemsFiltered,
     availableItemsFiltered,
     searchText,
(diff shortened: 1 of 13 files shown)


vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
```diff
@@ -0,0 +1,309 @@
...
+  Input,
+  ListBox,
+  ListBoxProps,
+  Option,
+  OptionProps,
+} from "@salt-ds/core";
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { applyHighlighting } from "@vuu-ui/vuu-table";
+import { Icon, IconButton } from "@vuu-ui/vuu-ui-controls";
```
heswell: don't import from the library like this, you'll end up with circular references

just use a relative path import

vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
```diff
@@ -0,0 +1,309 @@
...
+import { DragDropProvider, useSortable } from "@vuu-ui/vuu-utils";
+import {
+  ItemDescriptor,
+  CreateCustomItemProps,
+  ItemPickerHookProps,
+  useItemPicker,
+} from "./useItemPicker";
+import cx from "clsx";
```
heswell: use the 'organize imports' command in VS Code to sort the import statements into standard  typescript order.

vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
```diff
@@ -0,0 +1,309 @@
...
+  itemTypeSingular: string;
+  createCustomItemProps?: CreateCustomItemProps;
+}
+
+const getItemLabel = (item: ItemDescriptor) => {
+  if (item.label) return item.label;
+  else return item.name;
+};
+
+const searchIcon = <span data-icon="search" />;
```
heswell: use the vuu Icon component  
```
<Icon name="search"/>
```

vuu-ui/packages/vuu-ui-controls/src/item-picker/useItemPicker.tsx
```diff
@@ -0,0 +1,150 @@
+import {
+  FormEventHandler,
+  MouseEventHandler,
+  useCallback,
+  useMemo,
+} from "react";
+import { reorderColumnItems as reorderItems } from "@vuu-ui/vuu-utils";
```
heswell: we should probably move reorderColumnItems to array-utils and rename it as reorderItems, needs care as it us used fromn table as well

vuu-ui/packages/vuu-ui-controls/src/item-picker/useItemPicker.tsx
```diff
@@ -0,0 +1,150 @@
...
+  }, []);
+
+  const handleAddItemToSelectedList = useCallback<
+    MouseEventHandler<HTMLButtonElement>
+  >(
+    (e) => {
+      const name = itemName(e.target);
+      const itemToAdd = allItems.find((item) => item.name === name);
+      if (itemToAdd) {
+        const newSelectedItems = [...selectedItems, itemToAdd];
```
heswell: this is not a big deal, but its generally a better practise to use concat rather than spreading the existing array, so

```
newSelectedItems = selectedItems.concat(ItemToAdd)
```

when you use [...existingArray], you are iterating over the existing array, concat is a more efficient opertation. for tiny arrays like this its not a big deal,  but we have concat for exactly this scenario

vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
```diff
@@ -0,0 +1,309 @@
...
+ * As input props, all available items, currently selected items, search text and callback functions for changes in selected items and search text
+ * must be supplied by the client.
+ */
+export const ItemPicker = forwardRef(function ItemPicker(
+  {
+    className,
+    itemTypeSingular,
+    allItems,
+    selectedItems,
+    searchPattern,
```
heswell: managing the search text should be internal to the Picker component. Its purely a means to the end of producing the next selectedItems list.

vuu-ui/showcase/src/examples/UiControls/ItemPicker.examples.tsx
```diff
@@ -0,0 +1,162 @@
...
+  itemTypeSingular: string;
+}
+
+const StatefulParent = ({
+  allItems,
+  initialSelectedItems,
+  itemTypeSingular,
+}: StatefulParentProps) => {
+  const [selectedItems, setSelectedItems] = useState(initialSelectedItems);
+  const [searchPattern, setSearchPattern] = useState("");
```
heswell: better to move searchPattern into the Picker hook.

vuu-ui/showcase/src/examples/UiControls/ItemPicker.examples.tsx
```diff
@@ -0,0 +1,162 @@
...
+          `${item.label ? item.label : item.name}${i < newSelectedItems.length - 1 ? ", " : ""}`,
+        );
+      }
+
+      setSelectedItems(newSelectedItems);
+    },
+    [],
+  );
+
+  const handleSearchPatternChange = useCallback((newSearchPattern: string) => {
```
heswell: there may come a day when we source the allItems list from remote vuu data and might want something like this to apply filtering, but I suggest we keep it simple for now and keep search within the Picker component.

vuu-ui/showcase/src/examples/UiControls/ItemPicker.examples.tsx
```diff
@@ -0,0 +1,162 @@
...
+    console.log(
+      `handleSearchPatternChange() called with new search pattern ${newSearchPattern}`,
+    );
+
+    setSearchPattern(newSearchPattern);
+  }, []);
+
+  const customItemProps: CreateCustomItemProps = {
```
heswell: allow this to be passed to StatefulParent as a prop and include an example that does not have a custom Button (helps us ensure styling is good in all seenarios)

vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
```diff
@@ -0,0 +1,309 @@
...
+      if (listItems) {
+        const orderedItemNames = Array.from(listItems).map(getOptionName);
+        onReorderSelectedItems(orderedItemNames);
+      }
+    }, 300);
+  }, [availableItemsFiltered, selectedItemsFiltered]);
+
+  const searchPlaceholderText = `Find ${itemTypeSingular.toLowerCase()}`;
+  const selectedItemsHeading = `${itemTypeSingular.charAt(0).toUpperCase()}${itemTypeSingular.toLowerCase().slice(1)}s in view`;
+  const availableItemsHeading = `Available ${itemTypeSingular.toLowerCase()}s`;
```
heswell: this won't work for every name - Entry/Entries

I suggest we use a type like this

itemTypeName: ItemTypeName

where 

```
ItemTypeName = string | {
  singular: string;
  plural: string;
 }
```

That way user can just pass a string if adding an 's' is good enough.
For a non-standard plural they can speficy both. 

Worth a custom function to pluralize. I'd suggest we add a pluralise function into vuu-utils and define the above type there, this is likely to be useful elsewhere.

vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
```diff
@@ -0,0 +1,309 @@
...
+        listRef.current?.querySelectorAll<HTMLDivElement>(".saltOption");
+      if (listItems) {
+        const orderedItemNames = Array.from(listItems).map(getOptionName);
+        onReorderSelectedItems(orderedItemNames);
+      }
+    }, 300);
+  }, [availableItemsFiltered, selectedItemsFiltered]);
+
+  const searchPlaceholderText = `Find ${itemTypeSingular.toLowerCase()}`;
+  const selectedItemsHeading = `${itemTypeSingular.charAt(0).toUpperCase()}${itemTypeSingular.toLowerCase().slice(1)}s in view`;
```
heswell: FYI vuu-utils/text-utils capitalize()

vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
Louisianna1: Yes, good point!

PR #2305, opened 2026-08-15
Clickhouse plugin should use parameterized queries

stonesmi:
Prevent SQL injection by moving from string concatenation to parameterised queries.

Also:
* Fixed a bug with typeahead where we weren't applying the users current filter before querying for distinct values.
* Fixed querying on ScaledDecimal and EpochTimestamp data types which weren't previously working.
* Reviewed our Jacoco code coverage levels.


plugin/clickhouse-plugin/src/main/scala/org/finos/vuu/plugin/clickhouse/provider/filter/ClickHouseFilterVisitor.scala
```diff
@@ -1,14 +1,32 @@
 package org.finos.vuu.plugin.clickhouse.provider.filter
```
stonesmi, line 1: This is where the bulk of the change is.

tanjjj (approved):

PR #2310, opened 2026-08-18
2231 restructure rpc handler

tanjjj:


vuu/src/main/scala/org/finos/vuu/net/rpc/DefaultRpcHandler.scala
```diff
@@ -1,46 +1,8 @@
...
 package org.finos.vuu.net.rpc
 
-import com.typesafe.scalalogging.StrictLogging
 import org.finos.vuu.core.module.typeahead.ViewportTypeAheadRpcHandler
-import org.finos.vuu.core.table.TableContainer
 
-import java.util.concurrent.ConcurrentHashMap
+trait DefaultRpcHandler() extends RpcHandler with ViewportTypeAheadRpcHandler {}
 
-class DefaultRpcHandler(implicit tableContainer: TableContainer) extends RpcHandler with StrictLogging {
```
stonesmi, line 10: can we make this class private and add a companion object that creates the impl under the covers?

stonesmi (approved):

PR #2314, opened 2026-08-20
#2312 add RowDataUtils

### LINKED ISSUE #2312: Add RowDataUtils
Add a row data utils class so that we can safely get vlaue of a given colum from RowData, or throw RowDataException

tanjjj:
closes #2312

### DIFF
--- a/vuu/src/main/scala/org/finos/vuu/core/table/util/RowDataUtils.scala
+++ b/vuu/src/main/scala/org/finos/vuu/core/table/util/RowDataUtils.scala
@@ -1,0 +1,43 @@
+package org.finos.vuu.core.table.util
+
+import org.finos.vuu.core.table.{RowData, RowWithData}
+
+import scala.runtime.RichLong
+
+object RowDataUtils {
+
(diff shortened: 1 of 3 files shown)


vuu/src/main/scala/org/finos/vuu/core/table/util/RowDataUtils.scala
```diff
@@ -0,0 +1,43 @@
+package org.finos.vuu.core.table.util
+
+import org.finos.vuu.core.table.{RowData, RowWithData}
+
+import scala.runtime.RichLong
+
+object RowDataUtils {
+
+  def getRequiredLong(rowData: RowData, columnName: String): Long = {
```
stonesmi, line 9: i think this will return 0, if the long is empty. we should test this from java code

tanjjj: added RowDataUtilsJavaTest

stonesmi (approved):

PR #2348, opened 2026-08-26
Show file errors in drop zone and add rowDefaults support

anandBalachandar:
**CSV upload error display (`CsvUpload.tsx`)**
- File-level validation errors (missing key column, unknown columns, max rows exceeded) are now listed in the drop zone under "Your file contains errors"
- CSS classes `vuuCsvUpload-errorList` and `vuuCsvUpload-errorItem` added for consumer styling

**Refactor**
- `MAX_ROWS_IN_CSV` moved from `csv-schema-validation.ts` to `csv-constants.ts` and exported

**Row defaults (`EditSession`, `useCsvUpload`, `useEditableTable`)**
- `EditSession` accepts an optional `rowDefaults`
- `rowDefaults` prop added to `CsvUploadProps` / `CsvUploadHookProps` and forwarded to `EditSession`
- `rowDefaults` prop added to `EditableTableHookProps`
- PW test added


vuu-ui/packages/vuu-data-editing/src/EditSession.tsx
```diff
@@ -85,6 +85,7 @@ export class EditSession extends EventEmitter<EditSessionEvents> {
   #cellCommitRevisions = new Map<string, Map<string, number>>();
   #deleteMode: DeleteRowMode;
   #editSessionApi: EditSessionApi;
+  #rowDefaults: Record<string, VuuRowDataItemType>;
```
heswell: might be an idea to give this type a name, like RowDefaultDataItemValues. Its used in several places and the name makes it extra explicit exactly what its for.

vuu-ui/packages/vuu-data-editing/src/EditSession.tsx
```diff
@@ -355,7 +358,7 @@ export class EditSession extends EventEmitter<EditSessionEvents> {
       throw Error("[EditSession] datasource does not support adding rows");
     }
 
-    const response = await addRow.call(this.dataSource, rowData);
+    const response = await addRow.call(this.dataSource, { ...this.#rowDefaults, ...rowData });
```
heswell, line 375: we should check that column names provided in the default values actually exist in the target table, else we'll get a server error.

vuu-ui/packages/vuu-data-editing/src/useEditableTable.ts
```diff
@@ -71,10 +73,13 @@ export const useEditableTable = ({
     }
   }, [VuuDataSource, columns, dataSourceProp, table]);
 
+  // Stabilise rowDefaults so an inline object literal doesn't recreate the session every render.
```
heswell: we do not normally add this check, we leave it callers responsibility to pass a stable reference. 

If for some reason user _is_ trying to pass a changed set of default values, it will silently not work  and may cause confusion. 

If they do that and it causes our code a problem by recreating the editsession, that is something we should look at too. 

We can always add a javadoic comment to say a stable reference is expected

vuu-ui/packages/vuu-data-editing/src/EditSession.tsx
heswell: the number of constructor parameters is now such that I would go for a typed object
EditSessionConstructorProps.

The only mandatory one being the dataSource

PR #2362, opened 2026-08-30
feat(vuu-utils): add CSV export utilities via session table subscription

anandBalachandar:
- Adds `export-utils.ts` to `vuu-utils` exporting three functions:
  - `exportToCsv(dataSource, copyOption, filename, excludeColumns, onError, onSuccess)` — creates an `"export"` session table via `createSessionDataSource`, then subscribes with a full-range request to drain all rows, serialises to CSV and triggers a browser download
  - `exportSessionTableToCsv(sessionDataSource, ...)`
  - `exportCsvTemplate(dataSource, filename, excludeColumns, columns?)` 
- Enforces a `MAX_EXPORT_ROWS = 10_000` limit; excess row count is reported via `onError` callback rather than a synchronous throw (the limit check fires inside the async subscription callback)
- Internal session columns (`vuuMsg`, `vuuAction`, `vuuRowNum`) are always stripped from output
- Adds `CsvExport.examples.tsx` showcase with four examples: export all rows, export selected rows, simple one-button export, and template download (all columns and named subset)

### DIFF
--- a/vuu-ui/packages/vuu-utils/src/export-utils.ts
+++ b/vuu-ui/packages/vuu-utils/src/export-utils.ts
@@ -1,0 +1,143 @@
+import type {
+  CopyOption,
+  DataSource,
+  DataSourceCallbackMessage,
+  DataSourceRow,
+  DataSourceSubscribedMessage,
+} from "@vuu-ui/vuu-data-types";
+import { metadataKeys } from "./column-utils";
+import { Range } from "./range-utils";
+
+const EXPORT_EXCLUDED_COLUMNS = new Set(["vuuMsg", "vuuAction", "vuuRowNum"]);
+const MAX_EXPORT_ROWS = 10_000;
+
+const csvCell = (value: unknown): string => {
+  const s = value == null ? "" : String(value);
+  return s.includes(",") || s.includes('"') || s.includes("\n")
+    ? `"${s.replace(/"/g, '""')}`
+    : s;
+};
+
+const triggerCsvDownload = (csv: string, filename: string): void => {
+  const blob = new Blob([csv], { type: "text/csv;charset=utf-8;" });
+  const url = URL.createObjectURL(blob);
+  const a = document.createElement("a");
+  a.href = url;
+  a.download = filename;
+  a.style.display = "none";
+  document.body.appendChild(a);
+  a.click();
+  document.body.removeChild(a);
+  // delay revoke to give the browser time to start the download
+  setTimeout(() => URL.revokeObjectURL(url), 10_000);
+};
+
+/** Downloads a single-row CSV containing only the column headers, for use as an import template. */
+export const exportCsvTemplate = (
+  dataSource: DataSource,
+  filename = "template.csv",
+  excludeColumns: string[] = [],
+  columns?: string[],
+): void => {
+  const schema = dataSource.tableSchema;
+  if (!schema) {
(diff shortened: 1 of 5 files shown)


vuu-ui/packages/vuu-utils/src/export-utils.ts
```diff
@@ -0,0 +1,143 @@
...
+          sessionDataSource.unsubscribe();
+          return;
+        }
+        if (totalSize > MAX_EXPORT_ROWS) {
+          sessionDataSource.unsubscribe();
+          onError?.(new Error(`exportToCsv: row count ${totalSize} exceeds the ${MAX_EXPORT_ROWS} row limit`));
+          return;
+        }
+        // request all rows in one shot
+        sessionDataSource.range = Range(0, totalSize);
```
anandBalachandar: Is this going to be an issue on the service side?

vuu-ui/packages/vuu-utils/src/export-utils.ts
```diff
@@ -0,0 +1,149 @@
...
+} from "@vuu-ui/vuu-data-types";
+import { metadataKeys } from "./column-utils";
+import { Range } from "./range-utils";
+
+const EXPORT_EXCLUDED_COLUMNS = new Set(["vuuMsg", "vuuAction", "vuuRowNum"]);
+const MAX_EXPORT_ROWS = 10_000;
+const CHUNK_SIZE = 1_000;
+
+const csvCell = (value: unknown): string => {
+  const s = value == null ? "" : String(value);
```
stonesmi, line 23: Are we formatting dates? Or just using the display value

stonesmi: Ideally we should use ISO-8601

eg. 2025-02-16T10:58:44.965071000Z

anandBalachandar: Thanks..yes, I was looking at injecting custom formatters.
Added now.

vuu-ui/packages/vuu-utils/src/export-utils.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+        if (totalSize > maxRows) {
+          sessionDataSource.unsubscribe();
+          onError?.(new Error(`exportToCsv: row count ${totalSize} exceeds the ${maxRows} row limit`));
+          return;
+        }
+        // request rows in chunks for predictable behaviour across local and remote DataSources
+        sessionDataSource.range = Range(0, Math.min(CHUNK_SIZE, totalSize));
+      } else if (message.mode === "batch" && message.rows) {
+        collectedRows.push(...message.rows);
```
heswell, line 126: Note: rows may be received from server in multiple batches and with no guaranteed ordering of batches. It is clients job to assemble them into correct order.

vuu-ui/packages/vuu-utils/src/export-utils.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+      }
+    }
+  };
+
+  sessionDataSource.subscribe({ range: Range(0, 0) }, handleMessage);
+}
+
+/**
+ * Creates an export session table from `dataSource` then streams all rows to a CSV download.
```
heswell, line 159: nit: there is no streaming, rows are all collected in memory, so only suitable for small download volumes

PR #2372, opened 2026-09-03
Support session tables whose schema diverges from the source view table

anandBalachandar:
## Summary

Editing, CSV import, and CSV export all create a server-side session table via
`createSessionDataSource`/`beginEditSession`. Until now the session datasource
was always built by inheriting the source (view) datasource's config — columns,
filter, groupBy, sort, aggregations — which breaks whenever the session table's
schema differs from the view table's (a dedicated edit table, a separate
import/export table, etc.): the session subscribes with columns it doesn't
have, and view-only filter/sort/groupBy get carried over incorrectly.

This PR adds an explicit `SessionDataSourceOverrides` mechanism so callers can
declare the session table's columns and expected `VuuTable` up front, and fixes
a related bug in the CSV export path where column overrides were silently
ignored for local (`ArrayDataSource`-backed) session tables.

## Changes

- **`SessionDataSourceOverrides`** (`vuu-data-types`): new `{ columns?, table? }`
  argument added to `EditApi.createSessionDataSource` / `beginEditSession`.
- **`sessionDataSourceConfig` / `assertExpectedSessionTable`** (`vuu-utils`):
  shared helpers — discard the view config's filter/groupBy/sort/aggregations
  when session columns are overridden; validate the server-assigned session
  table's module against the expected table.
- **`VuuDataSource` / `TickingArrayDataSource`**: `createSessionDataSource`
  (and `TickingArrayDataSource.beginEditSession`) now honour `overrides`.
  `VuuDataSource.beginEditSession` (legacy path) intentionally left unchanged —
  it's a dead/broken API independent of this change.
- **`EditSession` / `useEditableTable`**: new `editColumns`/`editTable` props.
  `rowDefaults` pruning moved out of `begin()` into `reconcileWithSessionSchema`,
  which now also warns on missing `editColumns` and discards pending edits on a
  key-column mismatch. Hook returns `editSchema`/`columnsDiverge` for consumers
  rendering a single `Table` across view/edit modes.
- **`CsvUpload`**: new `importSchema`/`importTable` props for uploading into a
  table whose schema differs from the target table. Schema is fetched via
  `getTableSchema` when only `importTable` is given.
- **CSV export (`export-utils.ts`)**: `exportToCsv`/`exportCsvTemplate`/
  `exportSessionTableToCsv` accept `overrides`. Fixed a bug where overrides were
  passed to `createSessionDataSource` but never to the subsequent `subscribe()`
  call, so local session tables ignored the column restriction entirely.
- **Docs**: `vuu-data-editing/DESIGN.md` (RPC routing, divergent edit tables),
  `vuu-table-extras/csv-upload/README.md`, new `vuu-utils/docs/csv-export.md`.

### DIFF
--- a/vuu-ui/packages/vuu-data-editing/DESIGN.md
+++ b/vuu-ui/packages/vuu-data-editing/DESIGN.md
@@ -67,38 +67,55 @@ default the session data source is built from the view data source config, which
 would carry view-only columns — and any filter, `groupBy`, `sort`, or
 aggregations that reference them — onto a table that does not have them.
 
-`editColumns` and `editTable` describe the edit table when it differs:
+`EditSession` and `useEditableTable` are deliberately unaware of this: neither
+accepts a schema, an expected table, or any other transport-specific
+configuration. `useEditableTable`'s contract is exactly:
 
 ```ts
 useEditableTable({
   dataSource: viewDataSource, // subscribed; carries the createSessionTable RPC
-  editTable: EDIT_TABLE, // stable VuuTable
-  editColumns: EDIT_COLUMNS, // stable string[]
   isEditMode,
   onCancel,
   onSave,
 });
 ```
 
-Both are forwarded to `createSessionDataSource` / `beginEditSession` as a
-`SessionDataSourceOverrides` argument. Given `columns`, the data source builds
-the session config from `sessionDataSourceConfig` rather than inheriting the
-view config; `table` is checked against the module of the server-assigned
-session table. The session table name itself is always server-generated.
+Divergence is instead configured where the view data source is constructed, via
+an opaque `session` property:
 
-These overrides only affect how the session data source is constructed. Which
-data source subsequent operations target is unchanged — see [RPC
+```ts
+new VuuDataSource({
+  table: VIEW_TABLE,
+  columns: VIEW_COLUMNS,
+  session: {
+    table: EDIT_TABLE, // stable VuuTable
+    columns: EDIT_COLUMNS, // stable string[]
+  },
+});
+```
+
+`createSessionDataSource` / `beginEditSession` apply this internally: given
+`session.columns`, the session config is built from `sessionDataSourceConfig`
+rather than inheriting the view config; `session.table` is checked against the
+module of the server-assigned session table. A call-time
+`SessionDataSourceOverrides` argument (used by `CsvUpload`/`exportToCsv`, whose
+target data source is often shared and not under the caller's construction
+control) always takes precedence over the data source's own `session` config
+when both are present. `EditSession.begin()` never supplies one — it calls
+`createSessionDataSource(copyOption, sessionType)` with no third argument, so
+for editing the effective overrides are always whatever the data source was
+constructed with. The session table name itself is always server-generated.
+
+This only affects how the session data source is constructed. Which data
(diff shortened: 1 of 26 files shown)


heswell:
I think these API changes leak session-data-source construction details into `EditSession` and `useEditableTable`, which should own edit lifecycle and routing rather than schema/transport configuration.

Could the divergent-session definition instead live on the source data source (or its factory/configuration), preserving the existing editing API? For example:

```ts
useEditableTable({
  dataSource: viewDataSource,
  isEditMode,
  onCancel,
  onSave,
});
```

```ts
await sourceDataSource.createSessionDataSource(copyOption, "edit");
```

The view data source could be configured with an opaque session target:

```ts
new VuuDataSource({
  table: VIEW_TABLE,
  columns: VIEW_COLUMNS,
  session: {
    expectedTable: EDIT_TABLE,
    columns: EDIT_COLUMNS,
  },
});
```

`VuuDataSource.createSessionDataSource` could then apply the target internally: subscribe to the edit columns, discard incompatible view query state, and validate the server-returned table. This keeps `EditSession` and `useEditableTable` independent of schemas, expected tables, and transport-specific concerns.

Longer term, it might be preferable for the create-session response to provide the session schema or a subscription descriptor, allowing the client to construct the session data source from server-authoritative metadata and eliminate both overrides entirely. That needs an RPC contract change; moving the current overrides into data-source configuration would preserve the protocol while removing the API leakage now.

anandBalachandar:
Good point, and agreed - pushed a change that moves the divergent-session definition onto the source datasource.

`EditSession`/`useEditableTable` no longer accept `editColumns`/`editTable` at all. `useEditableTable`'s contract is now exactly what you proposed:

```ts
useEditableTable({ dataSource: viewDataSource, isEditMode, onCancel, onSave });
```

and `EditSession.begin()` calls `sourceDataSource.createSessionDataSource(copyOption, sessionType)` with no third argument - same as your snippet.

The divergent target is instead configured on the datasource itself, via a new opaque `session` property:

```ts
new VuuDataSource({
  table: VIEW_TABLE,
  columns: VIEW_COLUMNS,
  session: { table: EDIT_TABLE, columns: EDIT_COLUMNS },
});
```

`createSessionDataSource`/`beginEditSession` apply it internally - same mechanics as before (`sessionDataSourceConfig` discards the view's filter/groupBy/sort/aggregations, the returned table's module is validated), just relocated. One deviation from your snippet: I reused the existing `SessionDataSourceOverrides` shape (`{ columns?, table? }`) for `session` rather than introducing a second `expectedTable`-named type, to avoid maintaining two near-identical shapes.

I kept the per-call override parameter on `createSessionDataSource`/`beginEditSession` itself - it's an existing part of the `EditApi` surface, not something this PR added - because `CsvUpload`/`exportToCsv` have a genuinely different constraint: their target datasource is often app-supplied and shared elsewhere, so it can't always be reconstructed with a fixed `session` config just for one feature. Rather than reintroduce that leakage into `EditSession` for their sake, `useCsvUpload` now builds a small local adapter (implementing only `createSessionDataSource`/`tableSchema`, the two members `EditSession` actually touches on the source datasource before a session is active) that injects the import override at the RPC call site. `EditSession` itself never uses the parameter anymore — for editing, the effective overrides are always whatever the datasource was constructed with.

PR #2391, opened 2026-09-07
Refactor ColumnPicker to use new ItemPicker internally

Louisianna1:
Refactor of the existing ColumnPicker component to use the recently introduced generic ItemPicker component internally. Overview of changes:

**ColumnModel Class**
- The #searchPattern private field and associated getter and setter and filtering logic have been removed since ItemPicker manages the search pattern state and filtering of selected items. A new private field #selectedColumnsFiltered and updateSelectedColumnsFiltered() function that accepts ColumnDescriptor[] parameter have instead been introduced, called by ItemPicker whenever the filtered selections change.
- The addItemToSelectedColumns() and reorderSelectedColumns() functions have both been replaced with an addRemoveOrReorderSelectedColumns() function that accepts a ColumnDescriptor[] argument representing the new selections since both of these functions were only called by ColumnPicker previously.  The addRemoveOrReorderSelectedColumns() function makes use of functions in array-utils to determine whether the change was column addition, removal or reordering and notify listeners as before.
- The existing removeItemFromSelectedColumns() function that takes the name of the column to be removed has been left in tact since it is used by the useTableConfig hook.

**ColumnPicker**
- The useColumnPicker hook now handles conversion between ColumnDescriptor and ItemDescriptor constructs.

**ItemPicker**
- In addition to the onSelectedItemsChange() input prop / callback, onSelectedItemsFilteredChange() has been introduced to notify the parent component when the set of selected items with filtering (search pattern) applied changes.


vuu-ui/packages/vuu-table-extras/src/calculated-column/CalculatedColumnPanel.tsx
```diff
@@ -50,7 +49,7 @@ export const CalculatedColumnPanel = ({
   );
 
   const suggestionProvider = useColumnExpressionSuggestionProvider({
-    columns: columnModel.selectedColumns,
+    columns: columnModel.selectedColumnsFiltered,
```
heswell, line 52: the original code was correct, the column expression editor needs to be able to pick from any of the columns in the view. We are on a completely different view here to the columnpicker, filtering is local to the columnpicker.

vuu-ui/packages/vuu-table-extras/src/column-picker/ColumnModel.ts
```diff
@@ -73,29 +80,11 @@ export type ColumnEvents = {
...
-  } else {
-    return columns;
-  }
-};
-
 export class ColumnModel extends EventEmitter<ColumnEvents> {
-  #searchPattern = "";
   #selectedColumns: readonly ColumnDescriptor[];
+
+  #selectedColumnsFiltered: readonly ColumnDescriptor[];
```
heswell, line 86: I think that now filtering is handled within the ItemPicker, the ColumnModel needn't care about it at all.

heswell (changes requested):
I haven't been through code carefully yet, I will do so within the next few days. Most serious issue is that two way editing of columns is broken - changes made via drag drop of Table Column Headers or Columns removed via ContextMenu in the table are no longer reflected in the ColumnPicker.

vuu-ui/packages/vuu-ui-controls/src/item-picker/ItemPicker.tsx
```diff
@@ -174,6 +175,7 @@ export const ItemPicker = forwardRef(function ItemPicker(
     selectedItems,
     maxSelections,
     onSelectedItemsChange,
+    onSelectedItemsFilteredChange,
     onSelectionChange,
     selected = NO_SELECTION,
     createCustomItemProps,
```
heswell: lets get rid of this prop and the button it causes to render. I think its a specific use case of ColumnPicker (and even then a niche case) .
Its confusing that we have a prop called 'createCustomItemProps' that gives no indication in the name of what it actually does, which is create a button in the UI.

We'll keep the picker focused on the task of picking list items.

vuu-ui/packages/vuu-table-extras/src/calculated-column/CalculatedColumnPanel.tsx
Louisianna1: As mentioned, I wanted to check this comment as the original implementation of the selectedColumns getter applied filtering:

  get selectedColumns() {
    return filterColumns(this.#selectedColumns, this.#searchPattern);
  }

Louisianna1: Similarly, I updated the navigateColumn() function of useColumnSettings hook to refer to the columnModel.selectedColumnsFiltered getter I introduced instead of columnModel.selectedColumns. Should this also work on the full set of columns ignoring filtering?

heswell: you will need to move all references to the searchPattern and all filtering operations into the ItemPicker. 
Yes navigateColumn needs to revert to selectedColumns, it wasn't correct to change to to selectedcolumnFiltered. It is used in a different view from the columnPicker view, and filtering has meaning only in the context of ColumnPicker

heswell (approved):

PR #2392, opened 2026-09-07
Support update mode in CSV export row collection

anandBalachandar:
Allow viewport-update messages with mode of "update" during exportSessionTableToCsv in addition to "batch" mode.

### DIFF
--- a/vuu-ui/packages/vuu-table-extras/src/csv-export/export-utils.ts
+++ b/vuu-ui/packages/vuu-table-extras/src/csv-export/export-utils.ts
@@ -354,7 +354,10 @@ export const exportSessionTableToCsv = async <TName extends string = string>(
           // request rows in chunks for predictable behaviour across local and remote DataSources
           nextRequestedFrom = Math.min(CHUNK_SIZE, totalSize);
           activeSessionDataSource.range = Range(0, nextRequestedFrom);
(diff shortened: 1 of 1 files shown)


vuu-ui/packages/vuu-table-extras/src/csv-export/export-utils.ts
```diff
@@ -354,7 +354,10 @@ export const exportSessionTableToCsv = async <TName extends string = string>(
           // request rows in chunks for predictable behaviour across local and remote DataSources
           nextRequestedFrom = Math.min(CHUNK_SIZE, totalSize);
           activeSessionDataSource.range = Range(0, nextRequestedFrom);
-        } else if (message.mode === "batch" && message.rows) {
+        } else if (
+          (message.mode === "batch" || message.mode === "update") &&
```
heswell, line 358: remove the message.mode check, these are the only two values it can have at this point.

It seems odd that you would get 'update' here - update denotes ticking data (or is supposed to) as opposed to a 'batch' result from something like  a range request or a filter request etc.

Getting an 'update' mode here points to an issue somewhere else. The whole 'update' vs 'batch' distinction was introduced to make AG Grid support more efficient for ticking data. Its technical debt now we have withdrawn support for ag grid.

PR #2394, opened 2026-09-07
Inherit server connection on session datasource creation

anandBalachandar:
Fixes an initialization issue in the CSV Upload flow where adding rows to a staged session table fails with rpcCall server or viewport are undefined exception.

### DIFF
--- a/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
+++ b/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
@@ -713,7 +713,6 @@ export class VuuDataSource extends BaseDataSource implements DataSourceBase {
         table: sessionTable,
         viewport: sessionTable.table,
       });
-      sessionDataSource.server = this.server;
       return sessionDataSource;
     } else {
       throw Error(
--- a/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
+++ b/vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
@@ -271,7 +271,7 @@ export class VuuDataSource extends BaseDataSource implements DataSourceBase {
 
   handleSessionMessageFromServer = (msg: DataSourceCallbackMessage) => {
     if (msg.type === "subscribed") {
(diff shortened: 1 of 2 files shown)


vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
```diff
@@ -708,11 +708,13 @@ export class VuuDataSource extends BaseDataSource implements DataSourceBase {
       // config this datasource was constructed with.
       const effectiveOverrides = overrides ?? this.#session;
       assertExpectedSessionTable(sessionTable, effectiveOverrides?.table);
-      return new VuuDataSource({
+      const sessionDataSource = new VuuDataSource({
         ...sessionDataSourceConfig(this.config, effectiveOverrides?.columns),
         table: sessionTable,
         viewport: sessionTable.table,
       });
+      sessionDataSource.server = this.server;
```
anandBalachandar: clients (like useCsvUpload) interact with temporary session tables exclusively via one-shot RPC writes (editSession.addRow()) rather than viewport subscriptions. Because they do not invoke .subscribe(), the sub-datasource’s inner this.server was never initialized, raising a fatal rpcCall server or viewport are undefined exception.

anandBalachandar: Took this out - handled in useCsvUpload - 'begin()' resolves asynchronously only after the 'subscribed' event is received, preventing any viewport identity or uninitialized server errors.

vuu-ui/packages/vuu-table-extras/src/csv-upload/useCsvUpload.ts
```diff
@@ -214,9 +236,15 @@ export const useCsvUpload = ({
         );
       }
       const currentSessionTable = sessionDataSource.table;
-      await editSession.end(save);
+      try {
+        await editSession.end(save);
+      } finally {
+        if (sessionDataSource.status !== "unsubscribed") {
```
heswell, line 242: you don't really need this check, the datasource already checks ststus

PR #2397, opened 2026-09-08
Resolve BigInt serialization issues on row additions and filter queries

anandBalachandar:
This PR resolves the uncaught `TypeError: Do not know how to serialize BigInt` thrown during standard `editSession.addRow` calls (occurring during inline additions and CSV upload processes).

Cell inputs typed as `long` were cast into native `BigInt` wrappers locally on the client-side inside `getTypedValue` to assert integer boundaries. When these values were serialized into stringified WebSocket payloads, standard `JSON.stringify` failed.

### DIFF
--- a/vuu-ui/packages/vuu-utils/src/filters/filterAsQuery.ts
+++ b/vuu-ui/packages/vuu-utils/src/filters/filterAsQuery.ts
@@ -4,6 +4,7 @@ import {
   SingleValueFilterClause,
 } from "@vuu-ui/vuu-filter-types";
 import { isDateTimeDataValue } from "../column-utils";
+import { stringIsValidLong } from "../data-utils";
 import {
   isMultiClauseFilter,
   isMultiValueFilter,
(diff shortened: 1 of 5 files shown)


vuu-ui/packages/vuu-utils/src/filters/filterAsQuery.ts
```diff
@@ -13,7 +14,9 @@ import { ScaledDecimal } from "../ScaledDecimal";
 
 const filterValue = (value: string | number | boolean | ScaledDecimal) =>
   typeof value === "string"
-    ? `"${value}"`
+    ? stringIsValidLong(value)
```
heswell: this will fix the BigInt issue, the only problem is that it will lose the quotes on string values that happen to be numeric.  Question then is - will the vuu server fail if it attempts to compare a numeric value with string column values. I imagine it will. 

The only surefire way to eliminate this is to make the column available to this code so it can check the type. We already have this as an optional parameter on FilterByQuery, maybe we need to make this mandatory.

opts?: { columnsByName?: ColumnDescriptorsByName }

PR #2403, opened 2026-09-10
Refactor edit-session counters

anandBalachandar:
- Replaced loop-recalculating #refreshEditCounts() #storeCellEdit
- Made all state count setters private
- deleteSelectedRows to rely on consumer-supplied selectedRowCount, removing unsupported key-based deleted collections
- Wires dataRow.vuuAction directly from UndoCellRenderer.tsx to undoRowChange()

### DIFF
--- a/vuu-ui/packages/vuu-table-extras/src/csv-upload/useCsvUpload.ts
+++ b/vuu-ui/packages/vuu-table-extras/src/csv-upload/useCsvUpload.ts
@@ -146,9 +146,7 @@ export const useCsvUpload = ({
   // EditSession's constructor takes no session/schema config - the override is applied
   // here, at the createSessionDataSource call site, since `dataSource` is supplied by the
   // caller and may be shared with a view that has no knowledge of the import table.
-  const importDataSource = useMemo<
-    DataSource & { tableSchema?: TableSchema }
-  >(() => {
+  const importDataSource = useMemo<DataSource>(() => {
     return {
       tableSchema: dataSource.tableSchema,
       createSessionDataSource: async (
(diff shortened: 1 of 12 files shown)


vuu-ui/packages/vuu-table-extras/src/csv-upload/useCsvUpload.ts
```diff
@@ -187,7 +191,7 @@ export const useCsvUpload = ({
 
         return sessionDs;
       },
-    };
+    } as unknown as DataSource & { tableSchema?: TableSchema };
```
heswell: why do we have to add tableSchema ? Its already on DataSource;

seems a bit surprising that we're creating something here and coercing it to be a DataSource, which it isn't.

Feels like we need a type for it.

vuu-ui/packages/vuu-table-extras/src/csv-upload/useCsvUpload.ts
```diff
@@ -146,11 +147,14 @@ export const useCsvUpload = ({
   // here, at the createSessionDataSource call site, since `dataSource` is supplied by the
   // caller and may be shared with a view that has no knowledge of the import table.
   const importDataSource = useMemo<
-    EditApi & { tableSchema?: TableSchema }
+    DataSource & { tableSchema?: TableSchema }
```
heswell: see comment below

vuu-ui/packages/vuu-table-extras/src/csv-upload/useCsvUpload.ts
anandBalachandar: good spot...missed removing it.

PR #2405, opened 2026-09-11
Session table import refinements

stonesmi:
* Provide the import rowKey to consumers, rather than forcing the UI to provide it.
* Remove stack traces from failed RPC call responses to the UI


heswell (approved):
lgtm

Issue #2415, opened 2026-09-11
I'd like an example project that shows the creation of tables and modules in python and then the starting of the view server.


chrisjstevo:
## Feature Request

Allow the ability to create the Vuu server from within a python environment. 

### Description of Problem:

This example allows us to use Vuu as a visualization tool for data within a python environment. It would allow us to create table definitions and add them to Vuu on start up, then to load information in the python environment to visualize in Vuu.

PR #2416, opened 2026-09-11
#2415 Added example python integration, including creating view serve…

chrisjstevo:
…r table and provider in Python.


example/python-integration/python/test_start_server.py
```diff
@@ -0,0 +1,74 @@
+"""
+Smoke test for start_server.py.
```
stonesmi, line 2: we should add a github action to keep testing this, or it will defo get broken

stonesmi (approved):

example/python-integration/python/test_start_server.py
chrisjstevo: I agree, will consider how best to do.

PR #2418, opened 2026-09-11
Changes to addRow and CsvUpload

anandBalachandar:
### Align addRow with RpcResult changes.

- Standardizes the returning payload of addRow in both the remote and test data sources to return a structured RpcResult. 
- This resolves critical edit session state-synchronization issues where server-side insertion failures incorrectly updated edit counters (addCount) and bypassed user-facing validation warning tooltips. Additionally, it adjusts the showcase examples to allow inline editing inputs and registers a simulated server-rejection path to test and demonstrate error bubbling. 

### CSVUpload changes.
- Simplify status evaluation inside CsvUpload component to support explicit "error", "success", or undefined states cleanly.
- Relocate {children} inside the FileDropZone container below the FileDropZoneTrigger.
- Update CsvUpload package README


vuu-ui/package.json
```diff
@@ -23,7 +23,7 @@
     "format:check": "biome check --skip=lint --skip=assist",
     "format": "biome format --write .",
     "build": "node ./scripts/build-all-rslib.ts",
-    "bump:versions": "node ./scripts/bump-versions.ts",
```
heswell, line 26: we don't use tsx anymore, we have moved node requirement to 24+ which natively supports typescript 

see "engines" in package.json

vuu-ui/packages/vuu-data-remote/src/VuuDataSource.ts
```diff
@@ -840,16 +840,18 @@ export class VuuDataSource extends BaseDataSource implements DataSourceBase {
 
   async addRow(
     rowData: Record<string, VuuRowDataItemType> = {},
-  ): Promise<true | string> {
+  ): Promise<RpcResultSuccess | RpcResultError> {
```
heswell: this is the same as simply `RpcResult`

vuu-ui/showcase/src/examples/DataTable/Tree.data.ts
```diff
@@ -2765,6 +2765,14 @@ export default [
               path: "src/examples/Table/InlineAddRow.examples.tsx",
             },
           },
+          {
```
heswell, line 2768: FYI: theres no need to keep this exactly the same as actual tree, its only used as a sample json file to render a Tree example

PR #2431, opened 2026-09-15
csv upload timing and dirty state

anandBalachandar:
#### Summary
Resolves a timing and sequence issue during the CSV import session initialization where consumer views initially rendered empty tables. Also introduces a new `isDirty` helper property on the `EditSession` class representing unsaved edits.


vuu-ui/packages/vuu-data-editing/src/EditButtons.tsx
```diff
@@ -27,9 +28,8 @@ export const EditButtons = ({
   onSave,
   saveLabel = "Save",
 }: EditButtonProps) => {
-  const [editState, setEditState] = useState<EditState>(
-    () => editSession?.editState ?? "clean",
-  );
+  const editState = useEditState(editSession);
+  const isDirty = editState !== "clean";
```
heswell: why not use editSession.isDirty here ?

